PR #3, opened 2016-02-10
Support generating Go code

bbarenblat:
When run with `-go`, generate code for the [RISC-V Go port](https://github.com/riscv/riscv-go).


parse-opcodes
```diff
@@ -4,6 +4,35 @@ import math
 import sys
 import tokenize
 
+# Instructions not used by the RISC-V Go port (and which should therefore not be
+# emitted).
+GO_UNUSED_INSTRUCTIONS = set((
```
colinschmidt: Is this necessary for the port? Can it just not use the extra instructions or is there some restriction at play?

parse-opcodes
```diff
@@ -767,6 +796,28 @@ def make_chisel():
   print '  }'
   print '}'
 
+def print_go_insn(name):
+  print '\tcase A%s:' % name.upper().replace('.', '')
+  # Yank bits 0-6 of the encoded instruction (i.e., the opcode) and print them.
+  print '\t\treturn 0x%x' % yank(match[name], 0, 7)
```
colinschmidt: How are you tracking the rest of the bits in the instructions (i.e. the funct fields)? Maybe I just don't understand how the port will use this file.

PR #4, opened 2016-03-18
Narrow the fence and fence.i masks to match the ISA specification.

michaeljclark:
Before change:

 #define MATCH_FENCE 0xf
 #define MASK_FENCE  0x707f
 #define MATCH_FENCE_I 0x100f
 #define MASK_FENCE_I  0x707f

After change:

 #define MATCH_FENCE 0xf
 #define MASK_FENCE  0xf00fffff
 #define MATCH_FENCE_I 0x100f
 #define MASK_FENCE_I  0xffffffff

### DIFF
--- a/opcodes
+++ b/opcodes
@@ -66,8 +66,8 @@ sh     imm12hi rs1 rs2 imm12lo 14..12=1 6..2=0x08 1..0=3
 sw     imm12hi rs1 rs2 imm12lo 14..12=2 6..2=0x08 1..0=3
 sd     imm12hi rs1 rs2 imm12lo 14..12=3 6..2=0x08 1..0=3
 
-fence       31..28=ignore pred succ     19..15=ignore 14..12=0 11..7=ignore 6..2=0x03 1..0=3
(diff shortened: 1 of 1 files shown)


michaeljclark:
Just noticed the mask for the fence instructions are a bit wide. Not sure of the reason...

aswaterman:
Mmm? The spec says that the unused fields should be ignored, for compatibility with future fine-grained fences.

michaeljclark:
OK. Can see that text now however the diagrams show zero values The patch leaves predecessor and successor open for "fence" and sets regs and immegiate to 0 for "fence.i" per the diagram.

What is the defined behaviour for calls with non-zero? This appears to be undefined behaviour.

If we apply this patch, the encoding space still remains free ... or does this have implications on the generated Chisel?

aswaterman:
The unused bits are only allowed to make the fences finer grain, not coarser, so if implementations disregard them, everything's kosher.  There's no undefined behavior.  I think the right thing to do is not merge the patch -- doing so would cause future finer-grain fences to trap, which we don't want.

michaeljclark:
OK. Understand. Closing the pull request.

PR #5, opened 2016-05-24
Add dret instruction and debug CSRs.

timsifive:


aswaterman:
I was just thinking it's a little weird to have the debug CSRs right in the middle of the machine R/W CSR space (0x700 - 0x7BF) -- might make more sense to put them in the 0x7Bx range.  Not that it really matters much.  What do you think?

aswaterman:
Of course, that would also mean changing DRET's opcode to 0x7B2xxxxx, to minimize decoding complexity.

timsifive:
I have no opinion on what the address of the CSRs should be, nor of the encoding of DRET. Let me know what you think is best and I will change it.

aswaterman:
In that case, let's do what I proposed.

timsifive:
OK, so: 0x700--0x702 for the CSRs, changing bits 31..20 of dret to 0x702. Does that sound right?

timsifive:
Actually, 700-702 are already in use. Do you prefer 707-709? 708-70a? 710-712?

aswaterman:
I meant 0x7Bx, not 0x70x.  The least-significant digits are fine as-is.

timsifive:
I've updated the code.

PR #9, opened 2016-08-14
Verilog RTL decoder generator.

ben-marshall:
Added a simple python script which reads in any of the opcode specification files and automatically generates the bulk of an instruction decoder.

It doesn't generate a complete model, rather, given a 32-bit input signal containing the raw instruction to be decoded, it will create a set of N wires, each of which are high iff the data on the input wire decodes to the instruction the wire represents. From this, one can easily select the appropriate parts of the instruction data and raise them to whatever output interface they are crafting. By way of example, running the script on the base opcode set generates code like this:

``` verilog
// .......
wire gen_bgeu =    (decode_in[14:12] == 3'd7)
                && (decode_in[6:2] == 5'd24)
                && (decode_in[1:0] == 2'd3);
wire gen_fadd_d =    (decode_in[31:27] == 5'd0)
                  && (decode_in[26:25] == 2'd1)
                  && (decode_in[6:2] == 5'd20)
                  && (decode_in[1:0] == 2'd3);
wire gen_sltiu =    (decode_in[14:12] == 3'd3)
                 && (decode_in[6:2] == 5'd4)
                 && (decode_in[1:0] == 2'd3);
// .......
```

In the interests of simplicity, no logic sharing is detected. Almost all synthesis tools are smart enough to do this automatically now.

One can also change the name of the input `decode_in` wire, and set a custom prefix for all generated wires, which is `gen_` by default.


ben-marshall:
Abandoning since this request is now so out of date with the master branch.

PR #10, opened 2016-08-22
Make hardware triggers match latest spec.

timsifive:


aswaterman:
updated per discussions today

encoding.h
```diff
@@ -68,6 +68,10 @@
 #define MCONTROL_STORE      (1<<1)
 #define MCONTROL_LOAD       (1<<0)
 
+#define MCONTROL_MASKMAX(xlen) (0x3f<<((xlen)-11))
```
timsifive: I wasn't sure what the convention for those was. This is a good solution. I'll change the spike code to use them.

timsifive:
Andrew's changes look good.

encoding.h
aswaterman: yeah, not sure there /was/ a convention...

aswaterman:
does this look right?

timsifive:
I think this captures our e-mail discussion and my spec updates.

aswaterman:
merged via 1a25601e9bbc2a0ab49203ef7a0b6d5a44c4b658

PR #18, opened 2017-12-26
Use old C style comments.

timsifive:
This improves the chance we can use this file with older, pickier
compilers.

Also it makes the OpenOCD patch check script happier.


timsifive:
Thank you.

PR #21, opened 2018-08-28
Add header following Go convention for generated code

tklauser:
Go has a convention for generated code comments
(https://golang.org/s/generatedcode), i.e. they must match the regex
^// Code generated .* DO NOT EDIT\.$

Adjust the generated header to follow this convention.


tklauser:
ping @aswaterman

PR #22, opened 2018-11-18
Modernize to Python 3

pavelkryukov:


aswaterman:
Does it still work with python2?

PR #23, opened 2019-01-26
Read opcodes from files

pavelkryukov:
Current generation flow assumes that opcodes are provided
in a `cat ./opcodes | ./parse_opcodes -c` manner. However,
Windows CMD has no `cat` command, and it uses `type' instead,
so implementation of cross-platform script is complicated.

In this patch, we allow parsing opcode files directly by Python,
if their names are provided as a command line arguments, not depending
on the host shell. If no arguments are passed, script behaves as usual,
reading opcodes from the stdin.

### DIFF
--- a/parse-opcodes
+++ b/parse-opcodes
@@ -928,70 +928,83 @@ def make_go():
   print('\treturn nil, false')
   print('}')
 
-for line in sys.stdin:
-  line = line.partition('#')
-  tokens = line[0].split()
+inputs = []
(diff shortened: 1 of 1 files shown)


pavelkryukov:
@palmer-dabbelt Could you please review this?

palmer-dabbelt:
I'm happy with the feature, but I'm going to punt to @aswaterman on the implementation as this is his repo :).

aswaterman:
Looks fine to me.

PR #24, opened 2019-02-11
Add SV Generation Capabilities

zarubaf:
This adds the capability to generate a System Verilog package file for RISC-V opcodes and CSR addresses.


aswaterman:
Lgtm

PR #25, opened 2019-04-23
Updated path to FESVR_H in Makefile

tovine:
Fixed the Makefile again after it was broken in this commit: https://github.com/riscv/riscv-tools/commit/1d08559a0280a0b975381ea4a0591ecd9ee56059


tovine:
Should also update the version of `riscv-opcodes` in `riscv-tools`, as it's currently pointing to a version that's over one year old.

PR #28, opened 2019-05-17
Rvv spec 0.7

chihminchao:
Based on v-spec 0.7

What have been done
    1. add required instruction encoding field
    2. add most of instruction encoding except for atomic extension

What are missing
    1. Some documentation generation based on legacy spec in v/rvv branch are not included and should 
        be added in the future

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,13 +6,13 @@ FESVR_H := ../riscv-isa-sim/fesvr/encoding.h
 ENV_H := ../riscv-tests/env/encoding.h
 OPENOCD_H := ../riscv-openocd/src/target/riscv/encoding.h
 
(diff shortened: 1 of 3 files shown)


aswaterman:
Great, thank you @chihminchao, I'll review tonight.

aswaterman:
Merging with some minor changes:

- Several instructions require the `vm` field to be 1.  Some of these are explicit in the spec, and others were missing from the spec; I have updated it.

- Added the instructions in the VMUNARY0 space, rather than having an instruction named VMUNARY0.

- Added some pseudoinstructions for the VMERGE instructions that help instruction decoding w.r.t. the constraint that vs2 must be 0 when the instruction is unmasked.

PR #31, opened 2019-08-04
(Partially) fix #30

### LINKED ISSUE #30: Hard to use as a module
The only way to use this is to modify the sources.  Two things would improve this:
- refactoring the code into a function so we can import it
- renaming parse-opcodes so it can be a legal module name.

tommythorn:
With this change (and a renamed parse-opcodes) it's possible to as a
Python module without having to patch the repo.  Example:
```
  from parse_opcodes import parse_inputs

  if __name__ == "__main__":
      (namelist, pseudos, mask, match, arguments) = parse_inputs(["opcodes", "opcodes-rvc"])
```


aswaterman:
Did you also mean to rename the file parse-opcodes to parse_opcodes?

tommythorn:
Wow you are fast.  I discovered immediately that I missed the add, but in time to push again you caught it :)

aswaterman:
Haha!  This looks good to me.  Thanks.

tommythorn:
Thanks!

aswaterman:
Actually @tommythorn this appears to be broken: ```waterman@Andrews-MacBook:~/riscv-opcodes$ cat opcodes opcodes-rvc-pseudo opcodes-rvc opcodes-custom opcodes-rvv | python ./parse_opcodes -c
Traceback (most recent call last):
  File "./parse_opcodes", line 1074, in <module>
    parse_inputs(sys.argv[1:])
  File "./parse_opcodes", line 999, in parse_inputs
    assert(0)
AssertionError
```

aswaterman:
It looks to me that we need to do something like this.  If you agree, I can commit it:

    diff --git a/parse_opcodes b/parse_opcodes
    index c9822d5..9ac586e 100755
    --- a/parse_opcodes
    +++ b/parse_opcodes
    @@ -1071,7 +1071,7 @@ def parse_inputs(args):
       return (namelist, pseudos, mask, match, arguments)
 
     if __name__ == "__main__":
    -  parse_inputs(sys.argv[1:])
    +  parse_inputs(sys.argv[2:])
 
       if sys.argv[1] == '-tex':
         make_latex_table()

tommythorn:
Thanks, please do.

PR #34, opened 2019-09-06
Update Go instruction encoding generation

4a6f656c:
Remove the 'ok' value since this is redundant - an invalid instruction can
already be identified by a nil return value.

Also, include the script flags in the 'DO NOT EDIT' header.


tklauser:
Friendly ping @aswaterman @chihminchao (you seem to be the ones committing to this repository most recently)

aswaterman:
@tklauser thanks for the ping

PR #36, opened 2019-11-28
Rvv 0.8 draft 20191118

chihminchao:
Follow the new spec change


aswaterman:
Thanks.  I also added vmv1r/vmv2r/vmv4r/vmv8r (and a pseudo for vmvnfr), removed vamo*q, and replaced vamo*d with vamo*e.

PR #38, opened 2020-03-30
Using OrderedDict to keep encodings for match/mask.

kito-cheng:
The purpose of this patch is prevent got different result in different python env.

I got lots unnecessary diff during updating encoding.h in riscv-pk, lots diff are just in different order,  and I realized the problem is cause by the dict type didn't guarantee the order when traversal items, so OrderedDict can prevent this problem in future I think.

### DIFF
--- a/parse_opcodes
+++ b/parse_opcodes
@@ -6,10 +6,11 @@ from builtins import range
 import math
 import sys
 import tokenize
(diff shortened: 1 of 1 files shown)


aswaterman:
Good idea.

PR #39, opened 2020-04-01
update csr to draft-20200326-af69f79

chihminchao:
about two parts
   1. hypervisor related
   2. missing high part for rv32

Signed-off-by: Chih-Min Chao <chihmin.chao@sifive.com>

### DIFF
--- a/parse_opcodes
+++ b/parse_opcodes
@@ -148,8 +148,16 @@ csrs = [
   (0x600, 'hstatus'),
   (0x602, 'hedeleg'),
   (0x603, 'hideleg'),
+  (0x604, 'hie'),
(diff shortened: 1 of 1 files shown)


parse_opcodes
```diff
@@ -208,7 +218,8 @@ csrs = [
   (0x7a3, 'tdata3'),
   (0x7b0, 'dcsr'),
   (0x7b1, 'dpc'),
-  (0x7b2, 'dscratch'),
+  (0x7b2, 'dscratch0'),
```
aswaterman, line 221: I feel like this change might cause some issues.  Make sure Spike still builds correctly.

PR #40, opened 2020-04-07
Add RV128I opcodes

pavelkryukov:


aswaterman:
Given how speculative RV128 is at this time, I'd prefer not to merge this PR.  Others may disagree.

Separately, it looks like the LaTeX changes delete a column of zeros from the shift amount without amending the headers.  We should just back out the LaTeX changes for now if we do end up merging the PR.

pavelkryukov:
How can we test LaTeX output? I could not find an instruction.

pavelkryukov:
> Given how speculative RV128 is at this time, 

By the way, RV128C and RC128Q are presented in the repository already.

aswaterman:
Yeah, I am also questioning whether they belong in the repo at this stage...

pavelkryukov:
Is it possible put all of them into a branch, similarly to `riscv-bitmanip` branch?

aswaterman:
I think that's a good idea.  I created a branch for rv128.  I'll close this PR, and you can open a new PR to that branch.

But before you do so, can you fix the LaTeX generation?  if you run `make instr-table.tex` on master, it should produce a bitwise-identical file before and after your changes.  In the current form of the PR, the shift instructions don't print correctly.

pavelkryukov:
Yes, sure.

pavelkryukov:
> In the current form of the PR, the shift instructions don't print correctly.

I see the only difference is the first column with one `0` bit less.
Shouldn't it be intended, as `shamt` might be 7 bits for RV128?

pavelkryukov:
Oh, I see, it says "RV64I". Ok, I'll check what can be fixed here.

PR #41, opened 2020-04-08
Add RV128I opcodes

pavelkryukov:
TeX output is not changed


aswaterman:
Thanks for fixing that, @pavelkryukov

PR #45, opened 2020-07-31
Rvv and hypervisor

chihminchao:


parse_opcodes
```diff
@@ -59,12 +59,16 @@ causes = [
   (0x06, 'misaligned store'),
   (0x07, 'store access'),
   (0x08, 'user_ecall'),
-  (0x09, 'supervisor_ecall'),
-  (0x0A, 'hypervisor_ecall'),
+  (0x09, 'hypervisor_ecall'),
+  (0x0A, 'supervisor_ecall'),
```
aswaterman, line 63: 0x09 should remain "supervisor_ecall" and 0x0A should be renamed "virtual_supervisor_ecall".

chihminchao: hmmm, the spike part has been changed like above and it should be fix.

aswaterman: Indeed.  It'll make the Spike code easier to read, though: it would be quite confusing if a guest OS could throw a "hypervisor_ecall" exception...

PR #52, opened 2020-11-07
Support generating Rust code

wusyong:
Add support for generating Rust code. People can generating the encodings via `make inst.rs`.


aswaterman:
lgtm, thanks

PR #54, opened 2020-11-14
Update RV128 branch

pavelkryukov:
Pulling latest changes to rv128 branch.
RV32B insturctions are adjusted to rv128 shamt size.
Not sure how to implement `fsri` in current environment. Currently, there is no way to say "this sequence must not be decoded in RV128 in that way".


pavelkryukov:
> Not sure how to implement `fsri` in current environment. 

Maybe, we can use "ignore" instead of "shamt", similar to RVC implementations.

PR #57, opened 2021-01-22
Drop bool return value in generated Go code

tklauser:
The `ok bool` return value is redundant and the version included in the
Go compiler doesn't use it, see
https://golang.org/src/cmd/internal/obj/riscv/inst.go. Thus, drop the
return value from the generating `parse_opcodes`a script.


tklauser:
#34 was already submitted with the same fix. Closing this PR.

PR #60, opened 2021-02-02
Add RISC-V P Extension v0.9.2 opcodes

hope51607:
Base on P-Extension Proposal v0.9.2 ([link](https://github.com/riscv/riscv-p-spec/blob/master/P-ext-proposal.adoc)).


aswaterman:
Thanks for keeping this up to date.

This design still uses the `1111111` major opcode, which is reserved for >32-bit instructions, so is not valid to use for the P extension.  We will wait to merge until that issue is resolved through RVIA.

hope51607:
@aswaterman
I have updated to p-ext v0.9.2 which change major opcode "GE80B 1111111" to "OP-P 1110111".

hope51607:
Hi, @aswaterman 
Is there any problem in this PR?

aswaterman:
@hope51607 Sorry for losing track of this.  I'm manually merging it to fix the conflicts.

Issue #47, opened 2020-08-19
Why shamt in this repo is [25:20]


Phantom1003:
https://github.com/riscv/riscv-opcodes/blob/8d59eaae6d42e5dffa71846eb8296a8a5a461994/parse_opcodes#L36

Hi, I found that the opcodes in above line is [25:20], which should be [24:20] in spec. Is there is a miswrite, or I miss something ?

2henwei:
It means bit24-20 in Python

Phantom1003:
> It means bit24-20 in Python

No, I do not think so, rd is defined:
```python
arglut['rd'] = (11,7)
```
[11:7] follows description in spec
![image](https://user-images.githubusercontent.com/39120057/90650422-71198800-e26e-11ea-83e2-f91f4b2f50c2.png)

kito-cheng:
RV64 uses the bit 25 as the 6-th bit for shamt.

[image: image.png]

kito-cheng:
![image](https://user-images.githubusercontent.com/2723185/90662734-500b6400-e27b-11ea-9312-6534f0676c10.png)

Phantom1003:
Got it, thanks, but I found that 32i also use the 6-bit shamt, may be there is a bug ?

aswaterman:
There's an alias for the RV32I version of SLLI: https://github.com/riscv/riscv-opcodes/blob/8d59eaae6d42e5dffa71846eb8296a8a5a461994/opcodes-pseudo#L2

Note that one uses `shamtw`, which uses bits 24:20.

PR #62, opened 2021-02-07
Force bit 25 zero for RV32I

PENGUINLIONG:
As in Section 2.4 of the unprivileged specification ([20191213](https://github.com/riscv/riscv-isa-manual/releases/download/Ratified-IMAFDQC/riscv-spec-20191213.pdf)), `shamt` of immediate shift instructions should occupy the same bit range as `rs2` in register-only instructions, to have fixed value of either `0b0000000` or `0b0100000` in the most significant 7 bits. While in the opcode description in this repository, the value of bit 25 was not enforced. Is there an specific reason to have such discrepancy?

https://github.com/riscv/riscv-opcodes/blob/03be826f17faedcaee7f60223f402850e254df0a/opcodes-rv32i#L30

### DIFF
--- a/opcodes-rv32i
+++ b/opcodes-rv32i
@@ -22,12 +22,12 @@ lui     rd imm20 6..2=0x0D 1..0=3
 auipc   rd imm20 6..2=0x05 1..0=3
 
 addi    rd rs1 imm12           14..12=0 6..2=0x04 1..0=3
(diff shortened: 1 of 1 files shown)


PENGUINLIONG:
Just found this

https://github.com/riscv/riscv-opcodes/issues/47

PR #63, opened 2021-02-19
scalar-crypto: Add opcodes for RV32K, RV64K

ben-marshall:
Hi there

This PR does what it says on the tin, it also adds some support to the `parse_opcodes` script for allowing instructions to be decoded in the context of `XLEN`, particularly for the Spike `encoding.h` file. It relates to riscv/riscv-isa-sim#649. See the commit message for a little more information.

Note that these opcodes are *currently* going through the opcode and consistency review process, shortly behind Bitmanip.

While the Spike issue is open, this PR may need to be iterated on, since the best mechanism for enabling Spike (and other consumers of this information) to decode instructions in the context of `XLEN` is still ongoing. Scalar Crypto is the first extension to run into this issue, but likely won't be the last.

I opted for the simplest approach for tagging instructions as RV32/RV64 only - a manual list inside `parse_opcodes`. This could be dynamically populated in the future, but I wanted something simple and bullet proof for the first implementation.

As ever, happy to iterate if a better solution exists.

Thanks,
Ben

### DIFF
--- a/opcodes-rv32k
+++ b/opcodes-rv32k
@@ -1,0 +1,21 @@
+
+#
+# This file contains opcode specifications for the RISC-V
+# Scalar Cryptographic instruction set extension.
+# These instructions appear in _only_ RV32.
+# ------------------------------------------------------------
+
+# Scalar AES - RV32
+aes32esmi     rt     rs2 bs 11..7=0 29..25=0b11011 14..12=0 6..0=0x33
+aes32esi      rt     rs2 bs 11..7=0 29..25=0b11001 14..12=0 6..0=0x33
+aes32dsmi     rt     rs2 bs 11..7=0 29..25=0b11111 14..12=0 6..0=0x33
+aes32dsi      rt     rs2 bs 11..7=0 29..25=0b11101 14..12=0 6..0=0x33
+
+# Scalar SHA512 - RV32
+sha512sum0r   rd rs1 rs2    31..30=1 29..25=0b01000 14..12=0 6..0=0x33
+sha512sum1r   rd rs1 rs2    31..30=1 29..25=0b01001 14..12=0 6..0=0x33
+sha512sig0l   rd rs1 rs2    31..30=1 29..25=0b01010 14..12=0 6..0=0x33
+sha512sig0h   rd rs1 rs2    31..30=1 29..25=0b01110 14..12=0 6..0=0x33
+sha512sig1l   rd rs1 rs2    31..30=1 29..25=0b01011 14..12=0 6..0=0x33
+sha512sig1h   rd rs1 rs2    31..30=1 29..25=0b01111 14..12=0 6..0=0x33
+
--- a/opcodes-rv32k
+++ b/opcodes-rv32k
@@ -2,7 +2,7 @@
 #
 # This file contains opcode specifications for the RISC-V
 # Scalar Cryptographic instruction set extension.
-# These instructions appear in _only_ RV32.
+# These instructions appear _only_ in RV32.
 # ------------------------------------------------------------
 
 # Scalar AES - RV32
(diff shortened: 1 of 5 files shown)


opcodes-rv32k
```diff
@@ -0,0 +1,21 @@
+
+#
+# This file contains opcode specifications for the RISC-V
+# Scalar Cryptographic instruction set extension.
+# These instructions appear in _only_ RV32.
```
mwachs5, line 5:
```suggestion
# These instructions appear _only_ in RV32.
```

opcodes-rv64k
```diff
@@ -0,0 +1,22 @@
+
+#
+# This file contains opcode specifications for the RISC-V
+# Scalar Cryptographic instruction set extension.
+# These instructions appear in _only_ RV64.
```
mwachs5, line 5:
```suggestion
# These instructions appear _only_ in RV64.
```

parse_opcodes
```diff
@@ -295,6 +345,9 @@ csrs = [
   (0xF12, 'marchid'),
   (0xF13, 'mimpid'),
   (0xF14, 'mhartid'),
+  (0xF15, 'mentropy'), # crypto ext
+
+  (0x7A9, 'mnoise'),
```
mwachs5, line 350:
```suggestion
  (0xF15, 'mentropy'), # crypto ext
  (0x7A9, 'mnoise'),
```

ben-marshall:
Hi @mwachs5  - thanks for the suggestions, I've added them in. I haven't squashed them into one commit yet, I wanted to wait and do that once the broader changes are agreed upon by you / @aswaterman / others. I hope that's alright?

aswaterman:
I'm fine with the software changes.

I do have concerns about the `rt` encoding, though.  Writing `rs1` is going to be an implementation headache, given the base ISA has taken care that the destination is always in `rd`.  (It's not so much about the mux; it's also that superscalar implementations need `rd` early on to identify the dataflow, as do renamed implementations.)

I note that all of the instructions that use `rt` have 0 in the `rd` field, suggesting this decision might have been motivated by saving opcode space?  But R-type instructions take up so little encoding space that we can afford the redundancy of having both fields.

Even if there is some fundamental reason that `rd` and `rs1` must reference the same register, encoding it twice and having the implementation raise an exception if they are non-equal is probably preferable to moving the `rd` field around.

ben-marshall:
> I'm fine with the software changes.

Great, glad to hear it.

> I do have concerns about the `rt` encoding, though. [...] suggesting this decision might have been motivated by saving opcode space? 

Yes, this has come up in other places (most recently in riscv-crypto#75) for all the reasons you mentioned. Indeed we chose the current encoding to save opcode space, because the instructions are slightly more expensive than a normal R-type, because of the small 2-bit immediate they include. There is no reason the instructions need `rd==rs1`, rather, one *can* use the instructions this way to enable the opcode space saving. Our judgement at the time was that the opcode space saving was more important.

However, everyone involved I think would be happier to have separate `rd`/`rs1`. I will raise this with the TG. The only motivation for `rt` was opcode space, so if you of all people are saying this isn't such a concern, then we can change it to have distinct rd/rs1. I'm very glad you saw this before public review.

In terms of this specific PR, and riscv-isa-sim#649 - would you prefer to merge these in now, and accept the changes to the encodings later? Or wait until the encodings have been changed? My preference would be for the former, since merging now would make everything up to date and consistent with the latest v0.9.0 scalar crypto release. We can then change the encodings at once across spike/binutils/llvm/qemu/sail later on.

Many Thanks,
Ben

aswaterman:
@ben-marshall Cool, thanks for considering my feedback.  I'm OK with either plan, so I'll follow your suggestion to merge now and will await follow-on patches.

ben-marshall:
Thanks a lot :+1:

PR #76, opened 2021-07-18
Virtual memory updates

daniellustig:
* Add Svinval instructions
* Add PTE defines for Priv 1.12 and Svpbmt


aswaterman:
LGTM, thanks!

PR #77, opened 2021-07-23
scalar-crypto: post arch-review aes32* opcode change

ben-marshall:
- After questions from the architecture review and subsequent cryptography task group meetings, we have stopped overlapping the aes32* and aes64* instruction encodings.

- We've done this in the name of removing complexity, because opcode space is not as tight as we thought it was when we originally overlapped them.

- Change affected by updating the aes32* opcodes only.

- We anticipate that all of the scalar crypto opcodes could be improved in the future, but hope these are okay for now. The `parse_opcodes` script reports no collisions.

- Submitting this ahead of the actual spec release because it makes downstream spike/sail/bintuils updates easier.


aswaterman:
The same consideration affects the SHA512 instructions, right?  My sense is that we only get the complexity reduction if we root out all of the overlapping cases.

ben-marshall:
Oh Duh. You are quite right. I will add those changes to this PR this week.

ben-marshall:
Okay. Now I'm confused - after your comment I assumed to myself that yes, we had overlapped the SHA512 instructions (consistent with the AES* ones). Apparently we never have overlapped the SHA512 instructions? The RV32 SHA512 instructions have `func3=0`, and the RV64 ones have `func3=1`.

So, just to be safe, where did you read that we had overlapped them?

aswaterman:
I assumed you only added things to the `rv32_only` and `rv64_only` lists in `parse_opcodes` _because_ they overlapped, as that's what the comment above implies.

We should delete those data structures now, along with the code in `parse_opcodes` that uses them.  We can always dig up the old commits in the git history if we need to revive this strategy at some point in the future.

Similarly, Spike no longer needs to deal with the `DECLARE_RV32_ONLY` macros and `DECLARE_RV64_ONLY` macros: the implementations of the instructions already have the appropriate `require_rv32` or `require_rv64`, so trapping will happen as desired anyway.

ben-marshall:
> We should delete those data structures now, along with the code in parse_opcodes that uses them. We can always dig up the old commits in the git history if we need to revive this strategy at some point in the future.

Yeah, agreed. The latest commit to this PR removes them. I'll unpick them from Spike next.

aswaterman:
Cool.  Thanks again!

bulat242:
Is it possible to change other bit(-s) instead of 28-th bit's changing for aes32* instructions? This helps to keep decoding strategy for K-extension, when 28-th bit is always set and opcode is 0110011, that allows use less bits to distinguish K-extension and other extensions.

ben-marshall:
Hi @bulat242 - well, it looks like I forgot to update that appendix section of the spec, my mistake.

All of the opcodes may yet be improved (or the decode strategy updated) before and during public review. I'll make sure your question is recorded, but please do raise it again at public review. I have no doubt you and others will find better ways of encoding these instructions than I have.

Cheers,
Ben

bulat242:
Thank you for information! Okey, I will repeat it at public review.

All best,
Bulat

PR #84, opened 2021-11-01
Zbkx renames xperm.n and xperm.b as xperm4 and xperm8.

mjosaarinen:
The crossbar permutation instructions xperm4 and xperm8 were not included in 1.0 Bitmanip, but are specified as a part of the Zbkx extension in 1.0 Scalar Crypto. The instruction naming convention was changed during the arch review -- previously xperm.n and xperm.b. Also renamed the presently unused xperm.h and xperm.w to xperm16 and xperm32 for consistency.


mjosaarinen:
Also changes the name and address of the entropy source CSR as per Scalar Crypto Spec 1.0rc. Currently ("seed"), previously other names and also pseudoinstructions.

aswaterman:
LGTM.  (I didn't manually verify the opcodes against the spec, but I assume you got that right :))

PR #85, opened 2021-11-02
Update trigger CSRs per Debug Spec v1.0: mscontext, scontext, hcontext

JanMatCodasip:
This change updates the trigger-related CSRs. 

It is also needed for OpenOCD to compile after encoding.h is re-generated. @timsifive - FYI.


aswaterman:
@ernie-sifive can you please review?

ernie-sifive (approved):
Looks right to me.

PR #86, opened 2021-11-02
hypervisor MPV and GVA bits for mstatush

mjosaarinen:
These `mstatush` bits are defined in the privileged / hypervisor specification, and used by Spike master branch, but not defined in this repo.

### DIFF
--- a/encoding.h
+++ b/encoding.h
@@ -33,6 +33,8 @@
 
 #define MSTATUSH_SBE        0x00000010
 #define MSTATUSH_MBE        0x00000020
+#define MSTATUSH_GVA        0x00000040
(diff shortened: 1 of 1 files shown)


scottj97 (approved):

scottj97 (changes requested):
We have a bit of a mess now with riscv-isa-sim since the Krypto changes have already been added here in riscv-opcodes. Let me create a PR here from an older commit and try to update riscv-isa-sim so that its encoding.h is properly created from this repo. Then the Krypto changes will apply cleanly.

aswaterman:
Subsumed by #87

PR #87, opened 2021-11-02
Update encoding.h to match recent Spike changes

scottj97:
This PR adds a few commits to match changes that had been hand-made in Spike's copy of encoding.h, so that recent Krypto changes can be more cleanly applied to Spike.

@aswaterman please merge (not squash) so that I have a commit (70c120b5464d0a77491f07e5d66c4523d63a86a4) which closely matches Spike's current encoding.h. I will then submit a PR to Spike with a new encoding.h generated from this repo.

After that, [Spike's #846](https://github.com/riscv-software-src/riscv-isa-sim/pull/846) won't have so many diffs that are unrelated to Krypto.

This subsumes and obsoletes #86.


aswaterman:
Thanks for keeping them in sync.

PR #91, opened 2021-12-14
Add Zicbo{m,p,z} instructions

a4lg:
This commit adds instructions defined in RISC-V Base Cache Management Operation ISA Extensions, version 1.0.0-rc2 (now ratified).

https://github.com/riscv/riscv-CMOs

RFC: I'm not sure that new file name `opcodes-zicbo` and addition to `make_latex_table` didn't satisfy your requirements (Zicbo{m,z} instructions are tentatively here because they can be used in U-mode depending on configuration).


aswaterman:
Thanks.  We can rework the LaTeX table later if need be (but given that the specs are moving to asciidoc anyway, it might be moot).

PR #93, opened 2021-12-15
Make intermediate header file

a4lg:
I find that testing header file generation requires a bit of work (needs multiple external source trees *that we can destroy*) and I need a feature to generate test header file without external repositories.  Also, I found that encoding.h is not updated on some circumstances (when remote encoding.h is newer than source files; in other words, when we are actively working on a *remote* repository).

So, Instead of making (and depending on) multiple header files in external repositories, this commit makes single intermediate file and copies it to external directories on install.

This will...

*   Reduce CPU time (only slightly, though).
*   Enable making header file *without* cloning and destroying external repositories (`make encoding.out.h`).
*   Ensure that latest encoding.h is generated and copied on install, even after fresh checkout on an external source tree.


aswaterman:
Agreed, this is an improvement.

PR #94, opened 2021-12-16
Add new CSR bits defined in Privileged Spec version 1.12

a4lg:
# Request for Review

*   Is it the right place to add such CSR bits?
*   Many CSR bits (with almost the same role) share the same bit position (e.g. `menvcfg.FIOM`, `henvcfg.FIOM` and `senvcfg.FIOM`). Should we declare those as aliases?  
    e.g. `#define HENVCFG_FIOM MENVCFG_FIOM`

# Resolved Issue

*   Bit 63 of `henvcfg` is named `STCE` (`VSTCE` in ratified privileged spec 1.12 but it was a priv spec error)

# Description

This commit adds CSR bits defined in the privileged specification version 1.12.
This is a part of my work to implement following extensions in Spike:

*   Zicbo{m,z}
*   (part of) Zkr (access control by `mseccfg`)
*   (part of) Svpbmt (feature control by `menvcfg`/`henvcfg`)

It will also help implementing enhanced PMP (Smepmp) and S/VS-mode timer service (Sstc) extensions.


aswaterman:
@gfavor can you comment on the henvcfg.STCE vs. VSTCE naming disagreement?  The "V" prefix has been in the priv spec since August, but it doesn't seem quite right, since none of the other henvcfg bits in the priv spec have the "V" prefix.  And I guess the Sstc spec, which omits the "V" prefix, is what got ratified.  Should we just fix the priv spec?

gfavor:
Yes, the VSTCE bit name is inconsistent with the naming of all the other bits in henvcfg.  The Sstc spec that was ratified also uses STCE for the henvcfg bit name (for the preceding reason).  So this priv spec "typo" should be corrected.

aswaterman:
Done, it's now named STCE in the priv spec.

a4lg:
Thanks, the name `STCE` looks okay.
Now I removed the "RFC" status and request (relatively) normal review.

aswaterman:
Yeah, I think this is good to go.  Thanks.

PR #96, opened 2021-12-21
removed non-ratified RV64B opcodes

purushothaman-p:
The following Zbs instructions were removed from opcodes-rv64b file. 
`sbclriw`, `sbclrw`, `sbextw`, `sbinviw`, `sbinvw`, `sbsetiw`, `sbsetw`

Refer to the [issue on riscv-bitmanip](https://github.com/riscv/riscv-bitmanip/issues/97). Their encodings are redundant with I extension and did not become ratified. 

As a result, these instructions are not available in the latest release of [Bitmanip specification](https://github.com/riscv/riscv-bitmanip/releases/tag/1.0.0) (Version [1.0.0-38-g865e7a7](bitmanip-1.0.0-31-g2af7256.pdf)).


aswaterman:
I agree we should remove these because we’ve explicitly rejected them, but I’d prefer not to remove other unratified RVB opcodes just yet, until they’re explicitly rejected.

PR #99, opened 2022-01-20
Synchronize priv-instr-table.tex with the Manual

a4lg:
This commit roughly synchronizes privileged instruction table with the ISA Manual with slight instruction order modifications.

With this commit, all privileged and unprivileged (not compressed) instructions listed in the instruction tables can be generated from riscv-opcodes.

This is based on:

*   riscv/riscv-isa-manual: commit f30a5f6de685 ("Update chapters 2 and 7 for Hypervisor v0.6")
*   riscv/riscv-opcodes: commit 65af4131c26f ("Virtual memory updates (#76)")


aswaterman:
LGTM.

PR #101, opened 2022-02-13
Add orc.b instead of gorci

pavelkryukov:


aswaterman:
This needs to be done as a more general refactoring of the Zba, Zbb, etc. extensions, rather than as a one-off.  There are a few reasons for this: getting rid of `gorci` will break the Spike build; it's confusing to do this for `orc.b` but not for `rev8`; and this doesn't handle the fact that there's still no such thing as "the B extension".

I'll see if I can make progress on that tonight.

pavelkryukov:
Cool, thanks.

aswaterman:
Success.

pavelkryukov:
@aswaterman I see `orc.b` disappeared with new scheme, is it intended?

aswaterman:
I see `orc.b` still there, but as a pseudo.  This is OK with me.  https://github.com/riscv/riscv-opcodes/blob/9b9f61e34431f9924a0ce57b6cb8c61c63c9d97d/rv_zbb#L15

PR #105, opened 2022-04-02
Clean up handling of scalar crypto extensions

ZenithalHourlyRate:
Similar to 2497a8cc Clean up handling of bitmanip extensions

cc @ben-marshall


ben-marshall:
Only commenting since I'm tagged, but this seems like a very uncontroversial harmonising of filenames.

PR #107, opened 2022-04-16
add opcodes for zc extensions v0.70.3

liweiwei90:
The spec for zc extension(version 0.70.3) can be found in [zc spec](https://github.com/riscv/riscv-code-size-reduction/releases/tag/v0.70.3-TOOLCHAIN_DEV)


aswaterman:
Ack, but we won't merge this until it is clear the encodings won't change.

liweiwei90:
> Ack, but we won't merge this until it is clear the encodings won't change.

OK. I'll update this if the encodings changes

aswaterman:
Merging, since these are frozen.

aswaterman (approved):

PR #108, opened 2022-04-28
add spinalhdl support

Ncerzzk:
Signed-off-by: ncer <huangzzk@bupt.edu.cn>

Hi. I add the parse_code support for spinalhdl(https://github.com/SpinalHDL/SpinalHDL), which is similar with chisel,  so I just change encoding way of mark code from chisel's.

Please review this patch, thank you.


aswaterman:
Thanks for refactoring this to avoid code duplication.  (In future, we prefer that you squash your own commits when appropriate, as would be the case here.  That way, we can preserve important commits that are important to the history, without also preserving unimportant ones.)

aswaterman (commented):
(Forgot to explicitly approve)

PR #113, opened 2022-05-09
Dump masks to C output

pavelkryukov:
These masks may be used to write parsers


aswaterman:
I think ARG_LUT_ is not a great name prefix to put in an API. (For one, these are no longer lookup tables at this point.)

maybe something like RISCV_FIELD_? What do others think?

neelgala:
`RISCV_FIELDS_` or `RISCV_ARGS_` is probably better

pavelkryukov:
Other macros don’t have `RISCV_` prefix. Maybe `INSN_FIELD_`?

neelgala:
`INSN_FIELD_` sounds fine

PR #119, opened 2022-05-14
Test error inputs handling

pavelkryukov:


aswaterman:
OK with you, @neelgala?

parse.py
```diff
@@ -13,6 +13,11 @@
 pp = pprint.PrettyPrinter(indent=2)
 logging.basicConfig(level=logging.INFO, format='%(levelname)s:: %(message)s')
 
+def parse_constant(string):
+    base = 16 if 'x' in string else 2 if 'b' in string else 10
```
neelgala, line 17: this entire function can just be `return int(string,0)`  0 automatically indicates python to select the base based on the string prefix

neelgala:
made one comment.. all else looks good.. sorry for the delay - thanks @pavelkryukov for all the work here :D

parse.py
neelgala: @pavelkryukov here is my comment

PR #122, opened 2022-05-18
Check for overlaps between single bits and ranges

pavelkryukov:


neelgala:
@aswaterman please merge this after #124

Issue #125, opened 2022-06-02
HINTs are no longer emitted in encoding.out.h


aswaterman:
...which breaks Spike.  (If you copy the file into the HEAD of riscv-isa-sim master, you'll see what I mean.)

@neelgala can you resolve this one?

cf. https://github.com/riscv-software-src/riscv-isa-sim/issues/1020

aswaterman:
@neelgala can you resolve this one?

aswaterman:
related issue: https://github.com/riscv-software-src/riscv-isa-sim/issues/1020

neelgala:
I will look into this asap.

aswaterman:
Thank you!

aswaterman:
Resolved by https://github.com/riscv/riscv-opcodes/pull/126

PR #126, opened 2022-06-09
Explicit pseudo op inclusion

neelgala:
This should fix the issues highlighted in https://github.com/riscv-software-src/riscv-isa-sim/issues/1020 and #125


aswaterman:
Makes sense.  Thanks for following up on this, @neelgala.

aswaterman:
Resolved by https://github.com/riscv/riscv-opcodes/pull/126

PR #131, opened 2022-06-27
Include license directly into encoding.h.

timsifive:
This makes it much easier for programs that incorporate this code to
satisfy the following 2 license requirements:
1. Redistributions of source code must retain the above copyright
   notice, this list of conditions and the following disclaimer.
2. Redistributions in binary form must reproduce the above copyright
   notice, this list of conditions and the following disclaimer in the
   documentation and/or other materials provided with the distribution.

### DIFF
--- a/encoding.h
+++ b/encoding.h
@@ -1,7 +1,33 @@
-/* See LICENSE for license details. */
+/* See RISCV_ENCODING_LICENSE below for license details. */
 
 #ifndef RISCV_CSR_ENCODING_H
 #define RISCV_CSR_ENCODING_H
+
+#define RISCV_ENCODING_LICENSE \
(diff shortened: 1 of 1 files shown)


aswaterman:
I've never seen a license expressed as a macro before, so let's leave this open for a couple days to see if anyone has other thoughts.  Ping me again in a few days if nothing happens.  FWIW, I agree it's a reasonable way to facilitate binary distributions' reproduction of this license.

timsifive:
Another solution would be to use standard SPDX tags in the file to refer to well-known licenses. However, this license is almost but not quite identical to [BSD-3-Clause](https://opensource.org/licenses/BSD-3-Clause). I'm not a lawyer so I'm hesitant to say that they're functionally the same.

PR #135, opened 2022-07-09
Generalize `fence.tso` instruction

a4lg:
`fence.tso` instruction is encoded as a `FENCE` instruction with `fm=1000`, predecessor=`RW`, and successor=`RW` (as per the ISA Manual).

This commit generalizes `fence.tso` instruction to have unused *rs1* and *rd* operands (shall be zero on standard software).

In this PR, `fence.tso` is encoded like `fence` and `fence.i` instructions (that have generalized forms on riscv-opcodes).

## Not a Request: `pause` from `Zihintntl`

Note that `pause` instruction must not be edited like this because (quoting the ISA Manual),

> `PAUSE` is encoded as a `FENCE` instruction with *pred*=`W`, *succ*=`0`, *fm*=`0`, *rd*=`x0`, and *rs1*=`x0`.

### DIFF
--- a/rv_i
+++ b/rv_i
@@ -36,7 +36,7 @@ or      rd rs1 rs2 31..25=0  14..12=6 6..2=0x0C 1..0=3
 and     rd rs1 rs2 31..25=0  14..12=7 6..2=0x0C 1..0=3
 fence fm pred succ rs1 14..12=0 rd 6..2=0x03 1..0=3
 #specialized fences
-$pseudo_op rv_i::fence fence.tso 31..28=8 27..24=3 23..20=3 19..15=0 14..12=0 11..7=0 6..2=0x03 1..0=3
+$pseudo_op rv_i::fence fence.tso 31..28=8 27..24=3 23..20=3 rs1 14..12=0 rd 6..2=0x03 1..0=3
 $pseudo_op rv_i::fence pause     31..28=0 27..24=1 23..20=0 19..15=0      14..12=0 11..7=0      6..2=0x03 1..0=3
 ecall    31..20=0x000 19..7=0 6..2=0x1C 1..0=3
 ebreak   31..20=0x001 19..7=0 6..2=0x1C 1..0=3


neelgala:
The spec says : `The unused fields in the FENCE instructions—rs1 and rd—are reserved for finer-grain fences in
future extensions.`. So fence.tso should have those fields as 0 as per spec. The spec has reserved the unused encodings for future use.

a4lg:
I must say that `fence` and `fence.i` instructions have those reserved fields encoded as variables.

Although they are reserved,

> 2.7 Memory Ordering Instructions
>
> ...
> The unused fields in the `FENCE` instructions—*rs1* and *rd*—are reserved for finer-grain fences in future extensions.
> For forward compatibility, base implementations shall ignore these fields, and standard software shall zero these fields.
> Likewise, many *fm* and predecessor/successor set settings in Table 2.2 are also reserved for future use.
> Base implementations shall treat all such reserved configurations as normal fences with *fm*=`0000`, and standard software shall use only non-reserved configurations.

> Chapter 3: “Zifencei” Instruction-Fetch Fence, Version 2.0
>
> ...
> The unused fields in the `FENCE.I` instruction, *imm[11:0]*, *rs1*, and *rd*, are reserved for finer-grain fences in future extensions. For forward compatibility, base implementations shall ignore these fields, and standard software shall zero these fields.

it's not a bad idea to make this variable to safely ignore them without any illegal instruction fault (e.g. Spike is a base implementation and base implementations shall ignore those reserved fields/combinations).

neelgala:
hmm.. I see the inconsistency now (thanks).... and I am going to let @aswaterman decide which way to go - we can either accept this or fix fence/fence.i accordingly.

a4lg:
I don't recommend changing `fence` and `fence.i` since it will cause illegal instruction trap when reserved `fence` or `fence.i` is executed on Spike and it's against recommended behavior of RISC-V base implementations (as I quoted above). Because `fence.tso` is a subset of `fence`, not changing `fence.tso` is not harmful to Spike (despite that I think changing `fence.tso` is better).

So options I think is:

1. Change `fence.tso` as in this PR, or
2. Keep `fence`, `fence.i` and `fence.tso` as is.

aswaterman:
This is technically correct, even if surprising.

aswaterman (approved):

PR #137, opened 2022-07-21
Move the SPDX tag to the first line

borneoa:
Linux kernel coding style requires the SPDX tag to be placed in
the very first line of the generated file. See kernel file
 Documentation/process/license-rules.rst
at chapter 'License identifier syntax', '1. Placement:'.

Move the SPDX tag in the python script.
While there, adjust also the style of the multi-line comment.

Signed-off-by: Antonio Borneo <borneo.antonio@gmail.com>


aswaterman (approved):

timsifive (approved):
I think this is OK, assuming nobody will grab encoding.h by itself.

PR #138, opened 2022-08-12
add field definition for dcsr (debug v1.0.0)

YenHaoChen:
1. Add dcsr.debugver, dcsr.ebreakvs, dcsr.ebreakvu, dcsr.stepie,
dcsr.stopcount, dcsr.v, dcsr.mprven, and dcsr.nmip definitions.
2. Mark dcsr.xdebugver, dcsr.ndreset, dcsr.fullreset, dcsr.ebreakh,
dcsr.stopcycle, dcsr.debugint, and dcsr.halt deprecated. Those field
definition are kept for backward compatibility.

### DIFF
--- a/encoding.h
+++ b/encoding.h
@@ -63,18 +63,26 @@
 #define USTATUS_UIE         0x00000001
 #define USTATUS_UPIE        0x00000010
 
-#define DCSR_XDEBUGVER      (3U<<30)
-#define DCSR_NDRESET        (1<<29)
-#define DCSR_FULLRESET      (1<<28)
+#define DCSR_XDEBUGVER      (3U<<30)   // deprecated
+#define DCSR_NDRESET        (1<<29)    // deprecated
+#define DCSR_FULLRESET      (1<<28)    // deprecated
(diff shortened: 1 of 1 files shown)


aswaterman:
@YenHaoChen does this still match Spike?  Should I still merge it?

@timsifive OK with you?

YenHaoChen:
Spike needs a new encoding.h of this PR before implementing dcsr behavior in debug spec v1.0-STABLE.

aswaterman:
OK.  Will merge after @timsifive assents.

timsifive:
I'd ignore encoding.h altogether when it comes to the debug code, and use the macros defined in debug_defines.h. They're more complete, better documented, and auto-generated from the same source as the spec itself.

aswaterman:
@YenHaoChen would following Tim's advice make this PR unnecessary?

YenHaoChen:
Yes, debug_defines.h has a more complete definitions for debug code. I still propose to clean up those outdated definitions in encoding.h.

PR #140, opened 2022-08-22
Add copyright holder to the generated file

borneoa:
The BSD-3-Clause license mentions a copyright notice that has to
be retained while redistributing the source code.

Copy the copyright notice from LICENSE file to the script, so it
get in the generated file.

Signed-off-by: Antonio Borneo <borneo.antonio@gmail.com>


timsifive:
Works for me, assuming this is the blessed format.

aswaterman:
@timsifive

timsifive (approved):

PR #141, opened 2022-08-28
Make path to encoding.h relative to script

zarubaf:
That allows for running the script out-of-tree.

Signed-off-by: Florian Zaruba <florian@zaruba.email>


aswaterman (approved):
Good point.

PR #142, opened 2022-09-01
throw error when args contains a variety of base extensions

lhtin:
When use parser.py to generate encoding.out.h, the RV32I and RV64I SLLI instruction has the same marco name MATCH_SLLI and MASK_SLLI but diffrentent code encoding. So I think there need some check for this situation.

### DIFF
--- a/parse.py
+++ b/parse.py
@@ -133,6 +133,9 @@ def process_enc_line(line, ext):
 
     return (name, single_dict)
 
+def get_base_ext (ext_name):
+  return ext_name.split("_")[0]
+
 def same_base_ext (ext_name, ext_name_list):
     type1 = ext_name.split("_")[0]
     for ext_name1 in ext_name_list:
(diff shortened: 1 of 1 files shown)


neelgala:
This check is not required. The RV32I version of SLLI is the pseudo-op for the RV64I SLLI. So when generating the encoding.out.h the rv32i pseudo-ops will be ommitted (achieve by the second-arg [here](https://github.com/riscv/riscv-opcodes/blob/master/parse.py#L956)). 

some artifacts will require considering pseudo-ops as well for their generation (like the go artifact). So this has to be selected on a per artifact basis and the check is not required.

lhtin:
Oh, I think I missed the case that current spike will use the encoding.out.h with both rv32 and rv64 supported although I think it's odd that both rv32 and rv64 are present in the file. Previously I thought the user would only choose one of them (rv32 or rv64 or rv128) to use. Thank you for your reply.

PR #144, opened 2022-10-05
support import parse.py from other python script.

sequencer:
This PR changes opcodes_dir to the REAL path of parse.py. It helps python
scripts which depends on parse.py can be stored in other place, using PYTHONPAT
or sys.path to import from riscv-opcodes/parse.


neelgala:
Looks good to me.

aswaterman (approved):

PR #148, opened 2022-12-12
Removing import of csrrs and crrci in rv_zpn

eflaner:
Removing import of csrrs and crrci as the instruction's pseudo opcode is already included.

Signed-off-by: Babu P S <11073327+eflaner@users.noreply.github.com>


aswaterman:
What is the purpose of this PR, and what is the effect?

aswaterman:
Makes sense.  cc @neelgala

neelgala (approved):

PR #151, opened 2023-01-03
Fix pack instruction in unratified/rv_zbpbo

eflaner:
pack instruction should be imported from zbe -- instead of zbp. 

**Pack** instruction is currently defined in **unratified/zbe** 

Signed-off-by: Babu P S <11073327+eflaner@users.noreply.github.com>


neelgala:
This should have been caught by the CI scripts on the past PRs. Let me check whats broken there first.

neelgala (approved):

PR #155, opened 2023-01-14
Fix issues from #153

neelgala:
In #153 a couple of issues were introduced:
- the mnemonics of the shift-immediate ops in rv32 and rv128 (slli, srai and srli) were **replaced** with `_rv<xlen>` suffix - causing incompatibility to prior versions.
- the shift immediate ops in rv32_i were introduced as pseudo_ops of equivalents in rv128_i - which is wrong since in rv128_i the shift-immediates themselses are pseudo_ops of rv64_i
- the above caused a change in the Latex table for rv32i instruction names for shift ops by adding a suffix `.RV32`

This PR fixes the above issues by: 
- updating the checks in parse.py to ensure that pseudo ops are not dependent on other pseudo ops
- the intent of `PR#153` has been captured by **adding** the required pseudo ops instead of replacing the current ones.
- Updated the README on restrictions on usage of pseudo and imported ops

I have confirmed that the outputs for encoding.out.h remain the same as those from #153

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,15 @@ Instruction syntaxes used in this project are broadly categorized into three:
   ```
   $import rv32_zkne::aes32esmi
   ```
+
+### RESTRICTIONS
+
+Following are the restrictions one should keep in mind while defining $pseudo\_ops and $imported\_ops
+
+- Pseudo-op or already imported instructions cannot be imported again in another file. One should
+  always import base-instructions only.
+- While defining a $pseudo\_op, the base-instruction itself cannot be a $pseudo\_op
+
 ## Flow for parse.py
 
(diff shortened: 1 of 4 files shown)


neelgala:
@aswaterman please review and merge

aswaterman:
@neelgala sorry for screwing this up, but why did #153 pass the CI checks if it was so broken?

aswaterman (approved):

neelgala:
the "pseudo_op depending on another pseudo_op" check was not implemented properly (the regex was not strict enough therefore allowing substrings to be matched).

aswaterman:
Understood.  Thanks for adding guard rails, and I apologize once again for the screwup.

PR #147, opened 2022-12-02
updating opcodes of riscv-p-spec - v0.9.11 (c3409c8)

eflaner:
- Reorganized 'p' into sub-extensions as [described here](https://github.com/riscv/riscv-p-spec/blob/c3409c8edb7df262cb3db0ff323077120e5f7f04/P-ext-proposal.adoc#5-p-extension-subsets) 
    - created zpn, zpfs & zbpbo (along with appropriate 32 & 64 variants)
    - removed rv_p, rv64_p ( according to RISC-V sub-extension policy -- it's invalid )
- zbpbo 
    - [Replaced bitrevi with grevi](https://github.com/riscv/riscv-p-spec/blob/c3409c8edb7df262cb3db0ff323077120e5f7f04/P-ext-proposal.adoc#bitrevi-bit-reverse-immediate)) 
    - [Replaced BPICK with CMIX](https://github.com/riscv/riscv-p-spec/blob/c3409c8edb7df262cb3db0ff323077120e5f7f04/P-ext-proposal.adoc#111-bpick-bit-wise-pick)
- zpsf (zpsfoperand)
    - created zpsf as [specified in specification](https://github.com/riscv/riscv-p-spec/blob/c3409c8edb7df262cb3db0ff323077120e5f7f04/P-ext-proposal.adoc#52-zpsfoperand)
    - [SMMUL replacement in RV32 with MULH](https://github.com/riscv/riscv-p-spec/blob/c3409c8edb7df262cb3db0ff323077120e5f7f04/P-ext-proposal.adoc#42-rv32m-extension)
    - [MAXW ](https://github.com/riscv/riscv-p-spec/blob/c3409c8edb7df262cb3db0ff323077120e5f7f04/P-ext-proposal.adoc#112-maxw-32-bit-signed-word-maximum) and [MINW ](https://github.com/riscv/riscv-p-spec/blob/c3409c8edb7df262cb3db0ff323077120e5f7f04/P-ext-proposal.adoc#113-minw-32-bit-signed-word-minimum) removed (replaced with MAX and MIN of Zbb respectively)
    - [Removed SWAP8](https://github.com/riscv/riscv-p-spec/blob/c3409c8edb7df262cb3db0ff323077120e5f7f04/P-ext-proposal.adoc#7134-swap8-swap-byte-within-halfword) (replaced with REV8.H)
[...]

PR #158, opened 2023-02-06
Temporarily revert #147

aswaterman:
As a general rule, we need to make sure that the generated `encoding.out.h` suffices to compile Spike.  Normally, that's not a concern: adding new instructions, CSRs, etc. is benign.  However, the removal of instructions, CSRs, etc. that Spike already knows about breaks the build, so needs to be coupled with a corresponding Spike PR.  So, such changes to this repo should remain in an unmerged PR until the corresponding Spike PR is ready to go.

This PR should be reverted as soon as the Spike build is fixed (ideally by fully updating Spike, but possibly by temporarily removing the feature from Spike).

cc @eflaner: sorry again for the inconvenience.

cc @scottj97: after merging this, HEAD of master will once again be compatible with Spike.


neelgala:
@aswaterman https://github.com/riscv-software-src/riscv-isa-sim/pull/1247 This fix on spike master should allow reverting this PR.

PR #160, opened 2023-02-10
Feature/improved latex table

tovine:
I've made some changes to the LaTeX table generation (even though future documentation is supposed to be going towards AsciiDoc it's still useful to be able to generate full instruction set listings in the classical format.

This pull request includes automatic extraction of which instruction format headers to print (based on the actual instructions in the list), so there's no longer any need to provide them manually.
It also handles breaking up tables when they expand past a page boundary so longer extensions can be printed without manually splitting them (no idea why LaTeX can't do that automatically though).

Here's an example 16-bit table generated by this (see commented Zc* extensions in the diff): 
image

And here's an example of multi-page table splitting:
image

### DIFF
--- a/parse.py
+++ b/parse.py
@@ -475,6 +475,13 @@ def make_latex_table():
     dataset_list.append((['64_zfh'],'RV64Zfh Standard Extension (in addition to RV32Zfh)', [], False))
     make_ext_latex_table(type_list, dataset_list, latex_file, 32, caption)
 
+    # Zc* extensions
+    # caption = '\\caption{Zc* extensions for code size reduction on RISC-V}'
+    # dataset_list = [(['_zcb', '64_zcb'], 'Zcb (basic compressed ops) extension', [], True)]
+    # dataset_list.append((['_zcmp'], 'Zcmp (push/pop) extension', [], True))
+    # dataset_list.append((['_zcmt'], 'Zcmt (table jump) extension', [], True))
+    # make_ext_latex_table(type_list, dataset_list, latex_file, 16, caption)
+
     ## The following is demo to show that Compressed instructions can also be
     # dumped in the same manner as above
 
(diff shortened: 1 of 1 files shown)


tovine:
Note that I've not yet removed the `type_list` parameter from the function or the calls.
I first wanted to minimize the number of only-peripherally-related changes to avoid unnecessary reviewer fatigue and focus on the actual changes, but I can clean it up before merging. :)

neelgala:
@tovine This definitely looks cleaner and better. I will need sometime to verify this at my end. 

However, does this still produce the same tables as currently seen in the riscv-isa-manual ? If not @aswaterman should indicate if this change is acceptable.

tovine:
> @tovine This definitely looks cleaner and better. I will need sometime to verify this at my end.
> 
> However, does this still produce the same tables as currently seen in the riscv-isa-manual ? If not @aswaterman should indicate if this change is acceptable.

It should be exactly the same if only the same dataset is used, yes.
Here's how the compressed generation looks after the latest commit (added bit positions to the top):
image

aswaterman (changes requested):
In the originals, we only have one instruction format header per page, which subsumes all formats used on that page.  This version has a second set of headers on one of the pages; see the screenshot.

Screen Shot 2023-02-13 at 3 57 35 PM

tovine:
Thanks for the feedback, a bit busy this week but will make sure to take a look at fixing it when that changes!

aswaterman:
If it turns out to be more difficult than expected to generate the original tables exactly, then it might not be worth pursuing, since (as you said) we'll be moving to asciidoc in the next few months anyway.  Your call, of course.

tovine:
Sorry, I've been too swamped lately and haven't really had time to work on this at all. And the generated tables work well enough for my use so I can't justify working more on it.
You can decide if you'd like to accept it as is or just leave it hanging as an alternative branch

PR #161, opened 2023-02-11
Add the capability to generate a csv file

rpsene:
Signed-off-by: Rafael Sene <rafael@riscv.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ default: everything
 
 .PHONY : everything
 everything:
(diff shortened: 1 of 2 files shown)


neelgala:
can you please add the `-csv` to the github actions python-app.yml ([here](https://github.com/riscv/riscv-opcodes/blob/master/.github/workflows/python-app.yml#L23)(

neelgala:
@rpsene just curious.. could you share the use case a csv version of this ?
The PR otherwise looks good. Once you update the github actions yml, we can merge this

PR #170, opened 2023-03-16
Add (unratified) Zfa extension

ptomsich:
Zfa adds 'missing' floating-point functionality:
 - fli.[hsdq]: floating-point load-immediate
 - fminm/fmaxm.[hsdq]: IEEE 754-2019 minimum and maximum
 - fround(nx)?.[hsdq]: IEEE 754-2019 roundToIntegral(Exact)?
 - fcvtmod.w.d: ECMAScript modular float->integer conversion
 - move instructions for XLEN < FLEN
 - comparisons that support quiet NaN inputs

This adds a surprisingly large number of files, as the instructions added with Zfa depend on context (e.g., .h-variants become available only if Zfh is present).


aswaterman (approved):
I sanity-checked the encodings against the spec.

PR #184, opened 2023-07-31
fmvp.{q,d}.x: rs2 is variable field.

### LINKED ISSUE #183: Wrong encoding of fmvp.d.x instruction
Seems like **fmvp.d.x** is not in line with specification. Here is what spec states:

```
For RV32 only, if the D extension is implemented, the FMVP.D.X instruction moves a double-precision
number from a pair of integer registers into a floating-point register. Integer registers rs1 and rs2
supply bits 31:0 and 63:32, respectively; the result is written to floating-point register rd. FMVP.D.X is
encoded in the OP-FP major opcode with funct3=0 and funct7=1011001

```

The meaning of the above is that instruction consumes three registers, **rs1**, **rs2** and **rd** but this is how it is defined in **unratified/rv32_d_zfa**

`fmvp.d.x     rd rs1 24..20=1 31..27=0x16 14..12=0 26..25=1 6..2=0x14 1..0=3
`

From the above it can be oserved that **fmvp.d.x** consumes **rd** and **rs1** while having **rs2=1** which is not in line with spec.

nrajovic:
This resolves #183 .


ptomsich (approved):
LGTM. Thanks for the fix.

PR #186, opened 2023-08-13
Svadu: Rename HADE to ADUE in *envcfg

ved-rivos:
Svadu support was previously added in PR #156 . The *envcfg bit name were recently changed from HADE to ADUE.


aswaterman (approved):
LGTM, thanks @ved-rivos.

PR #187, opened 2023-09-16
Add unratified Zimop and Zcmop extension instructions

ved-rivos:
The Zimop extension introduces the concept of instructions that may be operations (MOPs). MOPs are initially defined to perform no useful operation and raise no exceptions, but might be redefined by later extensions to perform some other action. The Zimop extension defines an encoding space for 40 MOPs.

The Zcmop introduces eight code points in the 16-bit encoding space are provided for MOPs. The Zcmop instructions serve the same purpose as Zimop.


ved-rivos:
@aswaterman

aswaterman (approved):
Thanks @ved-rivos, lgtm

PR #194, opened 2023-09-23
Add pseudo-instructions for Zimop/Zcmop

ved-rivos:
Add mop.r.N, mop.rr.N, and c.mop.N pseudo-instructions that match all instructions in that group.


unratified/rv_zimop
```diff
@@ -1,7 +1,9 @@
 # The Zimop extension defines 32 MOP instructions named mop.r.0 -- mop.r.31.
 # They are encoded as 1-00--0111--sssss100ddddd1110011, where - denotes an
 # available opcode bit, s denotes the rs1 field, and d denotes the rd field.
+# add a pseudo-inst mop.r.N that matches mop.r.*
 
+mop.r.N  rd rs1 31=1 29..28=0 25..22=7 14..12=4 6..2=0x1C 1..0=3
```
aswaterman, line 6: I'm surprised this works, given that the parser is supposed to issue an error when multiple instructions use the same opcode.  Don't you need to use the `$pseudo_op` directive?  If not, did you accidentally detect a bug in the parser?

ved-rivos: The parser did not complain. Likely it only looks for identical code points but not an overlapping code point. I could not use the pseudo_op directive as unlikely a true pseudo-op which requires a 1:1 map with an op, this a 1:N map.

ved-rivos: Maybe we can remove this change in opcodes. For spike for example, we can achieve the same by using MOP_R_0_MATCH as both match and mask.

aswaterman: I think we’ve found one bug and one deficiency in the parser. It should detect any form of overlap, not just exact matches, and we should be able to declare pseudos that overlap multiple instructions. That’s what the @ directive used to do in the old version of this script.

IMO we should use this version of the patch, and fix the bug later, but it would probably be good to get @neelgala’s input before merging.

ved-rivos: I was a bit worried about that people using this may get multi-hits but I now think thats unfounded as they would either use the _N version or the individual decode and not both. So agree that we should use this.

neelgala: I think this is the case where not all bits are defined for c.mop.N (definition of bits 8-10 is missing). will need sometime to see how to catch this (if not already being caught).

ved-rivos: That was the intention with c.mop.N - it should match all 8 code points corresponding to the 8 encodings of 8-10. That by itself should not be an issue. The issue is that there are also c.mop.0, c.mop.1, ..., c.mop.7 - was expecting that the fact that c.mop.N overshadows c.mop.0-7 would be flagged..

ved-rivos: I wrote this function to detect such overlaps:
```
    for (int i = 0; i < count; i++) {
        for (int j = 0; j < count; j++) {
            if ( i == j ) continue;
            if ( (patterns[i].match & patterns[i].mask & patterns[j].mask) ==
                 (patterns[j].match & patterns[i].mask & patterns[j].mask) ) {
                printf("%s:%s\n", patterns[i].name, patterns[j].name);
            }
        }
    }

```
Attaching a list that was flagged
[output.txt](https://github.com/riscv/riscv-opcodes/files/12714981/output.txt)

ved-rivos: > well even so you will need to define those 3 bits as something (either an arg or constants) when creating c.mop.N. You will need to update the arg_lut to include a new arg for this.

That is a good idea.

ved-rivos: Is this the right syntax to use:
```
c.mop.N c_mop_t 1..0=1 6..2=0 11=0 7=1 12=0 15..13=3
$pseudo_op rv_zcmop::c.mop.N c.mop.0 1..0=1 6..2=0 11..7=1  12=0 15..13=3

```
I added this to the `arg_lut`
`arg_lut['c_mop_t'] = (10,8)`

neelgala: https://github.com/riscv/riscv-opcodes/blob/e6b30228b5cf6ced1909e62c336c70048ac59243/constants.py#L733 - include the list of pseudo instructions here so that they are emitted for spike. 

And your syntax looks correct

ved-rivos:
Thanks. I have updated this PR. It looks correct to me but I will test it with spike later today and let you know.

unratified/rv_zimop
ved-rivos: Thanks. I have updated this PR. It looks correct to me but I will test it with spike later today and let you know.

> - include the list of pseudo instructions here so that they are emitted for spike.

May be a good addition to the readme.

ved-rivos: Updated with new mnemonics for c.mop

aswaterman (approved):
Cool, thanks @ved-rivos and @neelgala.

PR #195, opened 2023-09-24
split arg_lut, causes, csr, csr32 from constants.py

sequencer:


aswaterman (commented):
I agree this is an improvement, but I'd like @neelgala and others to weigh in before merging.

sequencer:
Basically, the reason of this PR is I'm drafting a cleaner parser in Scala: https://github.com/sequencer/rvdecoderdb, which need to make all necessary information machine readable. I personally don't want to parse python script, thus move those these metadata into csv files.

neelgala:
lgtm.. but you will need to resolve the conflicts to get this merged.

sequencer:
@neelgala Fixed :)

sequencer:
Can we get this in?

aswaterman (approved):

PR #196, opened 2023-09-24
Remove duplicate zimm

sequencer:
Spot by Scala compiler :)
> ```scala
> Instruction.scala:50:20: unreachable code
> [warn]     case "zimm" => (19, 15)
> ```


aswaterman (approved):

aswaterman:
Static compilation FTW

PR #197, opened 2023-09-27
split dret from rv_system to rv_debug

sequencer:
Because neither priv and unpriv isa mentions dret, and debug is lived in
a standalone spec, move dret out from rv_system, while putting it into
rv_debug


aswaterman (approved):
Reasonable.

neelgala:
shouldn't the extension be rv_sdext (as per the debug spec)

sequencer:
> shouldn't the extension be rv_sdext (as per the debug spec)

agree, I think `rv_sdext` should live as "Debug ISA" as part of
```
Machine ISA
Supervisor ISA
Svnapot Extension
Svpbmt Extension
Svinval Extension
Hypervisor ISA
```

neelgala:
@sequencer can you raise another PR to rename rv_debug to rv_sdext ?

sequencer:
Done in #198 cc @neelgala.

PR #199, opened 2023-10-04
Add support for zicfiss extension with v0.3.1

SuHo-llrr:
Add Zicfiss encoding for spike simulator.

This patch checkout from d752f19 which last commit of riscv spike master encoding.h using.

We should rebase it after some extension ratified ex. Zimop/Zcmop


aswaterman:
These will change prior to freeze, so let's not merge yet.  cc @ved-rivos

SuHo-llrr:
already changed

PR #201, opened 2023-10-10
Adding Zalasr

mehnadnerd:
Adding load-acquire/store-release. Note they are written here as `lb.` for the load-acquire byte (so `lb.aq` and `lb.aqrl`), I'm not sure that will work but it passes the tests here.


aswaterman:
I get the strategy, but I can guarantee it'll cause some confusion :-) Let's see if it works out in Spike, Binutils, etc.

mehnadnerd:
I've updated it so it now uses `lb.aq` and `sb.rl` as the "base" names. I've got it working in Spike (https://github.com/mehnadnerd/riscv-isa-sim/tree/zalasr). I'm planning to do LLVM support next.

aswaterman:
Thanks, Brendan. I’ll take a look at the Spike PR later this week.

aswaterman:
Thinking about this some more, maybe consider adding back the `lb.` versions with `aq` and `rl` as args, but as pseudo-ops?  Implementations will want to decode them that way and, by happenstance, treat the other reserved aq/rl combinations the same way that LR/SC currently do.  Having these generic versions in the decode table facilitates that implementation style.

mehnadnerd:
Hmm, I think that will still have the confusion you mentioned. I'm also worried that people might choose to ignore the acquire bit for load-acquire, which would ruin compatibility for a future load-release.

aswaterman:
Fair enough. Implementers are obviously still free to do what they will with a reserved encoding.

aswaterman (approved):

PR #202, opened 2023-10-16
Remove unratified `Zvamo` instructions from `rv_v`

a4lg:
If this is the only thing happening, I would have just moved those instructions to `unratified/rv_zvamo`.

The reason I didn't is, there is the `Zabha` extension (containing subword AMO instructions) in the fast track and 8-bit vector AMO instructions conflict with it.

| Instruction     | Source               | 31..27  | 26   | 25   | 24..20 | 19..15 | 14..12 | 11..7 | 6..0      |
| --------------- | -------------------- | ------- | ---- | ---- | ------ | ------ | ------ | ----- | --------- |
| `vamoswapei8.v` | `Zvamo` (draft)      | `00001` | `wd` | `vm` | `vs2`  | `vs1`  | `000`  | `vd`  | `0101111` |
| `amoswap.b`     | `Zabha` (fast track) | `00001` | `aq` | `rl` | `rs2`  | `rs1`  | `000`  | `rd`  | `0101111` |

Related: riscv/riscv-zabha#17


aswaterman (approved):
This change will break Spike, so please also make a PR that removes the feature from Spike.

aswaterman:
Thanks for handling the Spike issue.

Issue #204, opened 2023-10-26
Unclear grammar / Exceptions to instruction syntax


tomhepworth:
I am having a go at writing a parser and I notice that most instructions are described by the format in the README  

IE `<instruction name> <instruction args> <bit-encodings>`

However there seem to be exceptions to this eg in `rv64_i`:

```
sraiw   rd rs1 31..25=32 shamtw 14..12=5 6..2=0x06 1..0=3
```

where bit encodings appear to be mixed with instruction arguments.

Is this a mistake or am I missing something?

Parsers could be written a lot easier with a consistent format.

aswaterman:
The format is consistent; it's just that args and bit encodings may be intermixed.  The only requirement is that the name is the first token and the others come later.

tomhepworth:
Perhaps the README should be updated to reflect this, since (at least to me) it currently implies that they are distinct. 

Perhaps just something like `<instruction name> <instruction args / bit encodings>`

tomhepworth:
Or better: `<instruction name> <arguments>` with a note that `<argument> := <bit encoding> | <variable argument>`

aswaterman:
I agree, and your most recent version is nicely concise while being accurate.  If you PR it, I'll merge it.

tomhepworth:
On it :)

sequencer:
FWIW, I wrote a Scala decoder here: https://github.com/sequencer/rvdecoderdb

tomhepworth:
Nice, I am making one in Haskell

PR #205, opened 2023-10-26
Clarified syntax of regular instructions

tomhepworth:
See https://github.com/riscv/riscv-opcodes/issues/204

Before this change the text implied that bit encodings and variable arguments could not be mixed in the list of instruction arguments.


aswaterman (approved):
LGTM; thank you for improving the README.

PR #213, opened 2023-11-29
Fix incorrect behavior between $pseudo_op and $import

charlie-rivos:
The passes for parsing $pseudo_op and $import are currently swapped, resulting in incorrect behavior. Pseudo ops are defined as:

"The pseudo op is only added to the overall dictionary is the dependent instruction is not present in the dictionary, else its skipped."

However, this is not the case if the dependent instruction has been imported.

Take the following examples:

rv_pseudo
```
$pseudo_op rv_op::op pseudo rd rs1 rs2 bs 29..25=0b11000 14..12=0 6..0=0x33
```

rv_import
```
$import rv_op::op
```

rv_op
```
op rd rs1 rs2 bs 29..25=0b11000 14..12=0 6..0=0x33
```

If we parse rv_pseudo and rv_op, the output is the single instruction "op". This is expected, as op is the "dependent instruction" and thus the pseudo is not included.

If we instead parse rv_pseudo and rv_import, the output is instead "op" and "pseudo". However, he output in this case should be the same as if the instruction was directly included. The "pseudo" instruction should not be included.

This change simply swaps the pseudo_ops and import passes, causing the behavior in both of these cases to be consistent with each other.

### DIFF
--- a/parse.py
+++ b/parse.py
@@ -252,71 +252,7 @@ def create_inst_dict(file_filter, include_pseudo=False, include_pseudo_ops=[]):
                 # update the final dict with the instruction
                 instr_dict[name] = single_dict
 
-    # second pass if for pseudo instructions
-    logging.debug('Collecting pseudo instructions now')
-    for f in file_names:
-        logging.debug(f'Parsing File: {f} for pseudo_ops')
-        with open(f) as fp:
-            lines = (line.rstrip()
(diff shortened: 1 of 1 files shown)


aswaterman:
This one has gone stale (sorry) but was woken up by the codecov bot for some reason.  I'm closing it, but feel free to reopen it as needed.

PR #214, opened 2023-12-19
CHERI encodings and CSRs

tariqkurd-repo:
- first attempt at CSRs and encodings for CHERI
- note that lr/b etc will be a separate extension eventually


tariqkurd-repo:
This is my first attempt at reserving some encoding and CSR space for CHERI.
There are some limitations:

LR.B, LR.H, SC.B, SC.H need to be in a separate extension, but for use by CHERI software
All immediate values are 12-bit which is unlikely to be acceptable due to the encoding space requirement.
However, I'm interested in getting some feedback on the encodings in general, so I thought it was worth getting this PR ready sooner rather than later

aswaterman (changes requested):
Using this repo to reserve opcode space far in advance of an extension's proposal isn't standard practice, and this is an even more unusual corner case since it's really a new base ISA.  Still, I think it's helpful to leave this PR open so people can look at the concrete proposal.

tariqkurd-repo:
Our intention is to keep CHERI as compatible as possible with other extensions, and so tentatively reserving encoding space and CSR numbers help with that goal. When the TG gets going next year we'll make a more formal proposal.

tariqkurd-repo:
@aswaterman this branch builds OK - but the `cincoffsetimm` encoding conflicts with `slli`. The conflict doesn't seem to be detected as the imemdiate values are different widths. Is this a known limitation?

veselypeta:
LLVM also includes a mechanism for detecting encoding conflicts, but it too failed to identify this. Perhaps for the same reason stated by above.

tariqkurd-repo:
See: https://github.com/riscv/riscv-opcodes/pull/237

PR #218, opened 2023-12-23
Update mstateen0 fields

ved-rivos:
Bit 56 - control access to henvcfgh
Bit 55 - control access to small state added by Priv 1.14
              - controls access to srmcfg introduced by Ssqosid
(The bit 56 was initially intended for use by the counter delegation extension but was not pursued/approved.)
https://github.com/riscv/riscv-isa-manual/issues/1147


encoding.h
```diff
@@ -168,11 +168,14 @@
 #define MSTATEEN0_CS       0x00000001
 #define MSTATEEN0_FCSR     0x00000002
 #define MSTATEEN0_JVT      0x00000004
+#define MSTATEEN0_PRIV14   0x0080000000000000
```
aswaterman, line 171: I think 114 would be a little clearer.

Do we also need to reserve the corresponding bit in `hstateen`?  I think the answer is "no, we can cross that bridge later", but wanted to make sure you agree and that it wasn't an accidental omission.

ved-rivos: Updated to PRIV114.

I did not reserve the bit in `hstateen0` yet as there was not a state defined.

aswaterman (approved):

PR #226, opened 2024-02-08
Update CLIC CSR_xINTSTATUS addresses, add CSR_xINTTHRESH

dansmathers:
update CSR_MINTSTATUS, CSR_SINTSTATUS, CSR_UINTSTATUS addresses to read-only csr addresses and add CSR_MINTTHRESH, CSR_SINTTHRESH, CSR_UINTTHRESH csrs.

spec: https://github.com/riscv/riscv-fast-interrupt

FYI @kito-cheng @cmuellner for notifying both LLVM and GCC community


chihminchao:
the change matches the latest spec
image
image
image

sobuch:
Hi all, user mode interrupts were dropped from CLIC spec in the meantime (https://github.com/riscv/riscv-fast-interrupt/commit/53539a3186d673a7e4ccf84c89f2f6cb9d60fd59). That seems more in line with contents in this repository, as all mentions of the N extensions were removed previously (except sedeleg and sideleg CSRs, which should be probably also removed).

Would this patch be accepted, if updated to drop the suclic CSRs?

PR #234, opened 2024-02-23
Add definition of low-priority and high-priority RAS event from AIA

YenHaoChen:
Add definition of low-priority and high-priority RAS event from AIA.
image


aswaterman:
Let's change these to `RAS_LOW_PRIO` and `RAS_HIGH_PRIO`.  The rationale is that RAS is the important part and priority is the less-important part, so RAS should come first in the name.

YenHaoChen:
@aswaterman Change the name as suggested

aswaterman (approved):

PR #236, opened 2024-03-02
move opcode of zvb*

Lucas-Wye:


rv_zvbb
```diff
@@ -0,0 +1,23 @@
+# Zvbb - Vector Basic Bit-manipulation
+# https://github.com/riscv/riscv-crypto/blob/main/doc/vector/riscv-crypto-vector-zvbb.adoc
+
+vand.vv        31..26=0x01 vm vs2 vs1 14..12=0x0 vd 6..0=0x57
+vand.vx        31..26=0x01 vm vs2 rs1 14..12=0x4 vd 6..0=0x57
```
aswaterman, line 5: These two instructions are not part of Zvbb; they are part of the base V spec.

aswaterman (changes requested):
Zvbb support already exists in the `unratified` directory: https://github.com/riscv/riscv-opcodes/blob/f7c71810cd75d52f5d613d9f752646e82533bf09/unratified/rv_zvbb

It should be moved from the `unratified` directory to the root once it is ratified.

sequencer:
Zvk instructions are all ratified. Maybe move all of them out?

aswaterman (approved):

PR #238, opened 2024-03-09
Add unratified Ssdbltrp and Smdbltrp fields

ved-rivos:


aswaterman (approved):
Sorry this fell through the cracks. I assume these remain accurate.

PR #244, opened 2024-05-04
Mark Zabha ratified

ved-rivos:


aswaterman (approved):
@ved-rivos did you do a quick scan and see if there are other extensions we should move over to ratified? If you don’t have the time, I’ll try to do so next week.

ved-rivos:
@aswaterman - sent PR #246 - for Zihintntl, Zicond, Zfa, and Zvbc

PR #255, opened 2024-06-28
remove _ in causes.csv

sequencer:
I think these names should be aligned, not matter with `_` or not with `_`


aswaterman (approved):
I agree it's illogical to use spaces for some and underscores for others.

PR #260, opened 2024-07-15
Fix definition of mcontrol.action

YenHaoChen:
The mcontrol has 4 bits only instead of 6 bits in debug spec 1.0.0-rc3.

### DIFF
--- a/encoding.h
+++ b/encoding.h
@@ -111,7 +111,7 @@
 
 #define MCONTROL_SELECT     (1<<19)
 #define MCONTROL_TIMING     (1<<18)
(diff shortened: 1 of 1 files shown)


YenHaoChen:
I found a CSR RW bug in Spike and traced it back to this root cause. 

The definitions of mcontrol CSR is outdated. People may want to remove all of the mcontrol definitions in the encoding.h and use the ones in debug_defines.h. In such a case, I can provide additional fixes to Spike.

aswaterman (approved):
I'm OK with continuing to provide these definitions in `encoding.h`, but of course you are right that they need to be correct.

PR #262, opened 2024-07-17
Fix Missing Extensions in Pseudo Instructions

NyembeziIMG:
- Added code to extend the extension list of a pseudo instruction in the YAML output
- This ensures that all the extensions that a pseudo instruction can belong to are listed
- This is in line with what happens for regular and imported instructions
- No changes to all other output types

### DIFF
--- a/parse.py
+++ b/parse.py
@@ -336,6 +336,13 @@ def create_inst_dict(file_filter, include_pseudo=False, include_pseudo_ops=[]):
                 if name not in instr_dict:
                     instr_dict[name] = single_dict
                     logging.debug(f'        including pseudo_ops:{name}')
+                else:
+                    # if a pseudo instruction has already been added to the filtered
+                    # instruction dictionary but the extension is not in the current
+                    # list, add it
+                    ext_name = single_dict['extension']
+                    if ext_name not in instr_dict[name]['extension']:
+                        instr_dict[name]['extension'].extend(ext_name)
             else:
                 logging.debug(f'        Skipping pseudo_op {pseudo_inst} since original instruction {orig_inst} already selected in list')
 


NyembeziIMG:
Zip with instr_dict.yaml before and after attached

[fix_pseudo_missing_extensions.zip](https://github.com/user-attachments/files/16267170/fix_pseudo_missing_extensions.zip)

aswaterman (commented):
Can you explain more about the problem you're solving?  What's missing and why should it be included?

(The code looks fine, and thanks for making sure that no other outputs are affected.)

NyembeziIMG:
Many thanks for the quick feedback.

Before this change, not all the extensions that the pseudo instructions belong to were listed under the extension field in instr_dict.yaml
As an example for rev8, only rv64_zks was listed under the extension field:
`rev8:
  encoding: 011010111000-----101-----0010011
  extension:
  - rv64_zks
  mask: '0xfff0707f'
  match: '0x6b805013'
  variable_fields:
  - rd
  - rs1`

After this change, rev8 lists 5 extensions, i.e. rv64_zks/rv64_zkn/rv64_zk/rv64_zbkb/rv64_zbb:
`rev8:
  encoding: 011010111000-----101-----0010011
  extension:
  - rv64_zks
  - rv64_zkn
  - rv64_zk
  - rv64_zbkb
  - rv64_zbb
  mask: '0xfff0707f'
  match: '0x6b805013'
  variable_fields:
  - rd
  - rs1`

This would be consistent with regular and imported instructions, which list all the extensions that they belong to.

aswaterman (approved):
Thanks for the explanation, sgtm.

PR #264, opened 2024-07-21
Refactor and Optimization:: constants.py

IIITM-Jay:
### About
This PR deals with refactoring and code optimisation towards a portion of `constants.py` file . A helper function is created to read `csv` files and returns the list of required tuples.

### Added/ Modified Changes

1. **A helper function  `read_csv` is created:** It takes the filename(str) as an argument and returns the list of tuples as before.
2. **Avoided repetition:** codes being written to remove the duplicated functionality to achieve the same output as a return
3. **Comments are added** 
4. **Code Optimisation Techniques are used**

### Code Optimisation Technique used

1. **Use of List Comprehension:** It allows us to use a shorter syntax to generate a list. For example, take a look in the code written earlier:
```
causes = []
with open("causes.csv") as f:
    csv_reader = csv.reader(f, skipinitialspace=True)
    for row in csv_reader:
        causes.append((int(row[0], 0), row[1]))
```
The following can be easily observed from the above piece of code:
    a. `causes` need to be declared before appending to it.
    b. The loop takes two line of codes to generate the tuple

Using _List Comprehension_ the above can be rewritten to the below modified code:
```
with open("causes.csv") as f:
    csv_reader = csv.reader(f, skipinitialspace=True)
    causes = [(int(row[0], 0), row[1]) for row in csv_reader]
```

2. **Removal of Duplicated Codes:** The codes for reading `csv` files and returning a tuple in the same format were written multiple times doing the same functionality. For better **_readability and to improve efficiency_**, a helper function is created which can be reused to return the same output as before.


IIITM-Jay:
Hi @aswaterman, this a step towards participation in the **_RISC-V Mentorship Program for the Fall 2024_** regarding the concerned project  [Refactoring and Enhancing the RISC-V Opcodes Repository](https://riscv.org/job/refactoring-and-enhancing-the-risc-v-opcodes-repository-risc-v-mentorship/). 
Requesting review from @aswaterman your side and for the feedback as well on what steps ahead to be taken for the above stated program. I have also seen the coding challenge mentioned on LFX platform. Will love to hear your views and opinions on the same.

Timmmm (approved):
LGTM. Minor thing: I would just suggest calling it `read_csv_int_map` instead because it reads a *specific* kind of CSV, not just any CSV file.

IIITM-Jay:
@aswaterman can these approaches be incorporated to reduce the number of lines of codes and to avoid redundancy. Let me know your thoughts on this PR. Will include the feedback and suggestions as provided. Thanks!

aswaterman (approved):
@IIITM-Jay Yes, LGTM.

Sorry that I'm very far behind on code reviews.  Please feel free to ping me on any high-priority PRs that have languished.

PR #271, opened 2024-08-01
Add RVV segment load/store instructions

wangpc-pp:
This is needed by golang (and maybe other languages) porting.

We set `nf` field explicitly.


wangpc-pp:
cc @4a6f656c

sequencer:
See https://github.com/riscv/riscv-opcodes/blob/07b21cc5143a15959eda12e30aa40cea0971efe0/arg_lut.csv#L45

wangpc-pp:
> See
> 
> https://github.com/riscv/riscv-opcodes/blob/07b21cc5143a15959eda12e30aa40cea0971efe0/arg_lut.csv#L45

I don't get it, what do you mean here? We should remove it from this table?

sequencer:
`nf` is already encoded in `arg_lut`, explicit using them in the table is strange, and may break the backward compatibility.

wangpc-pp:
> `nf` is already encoded in `arg_lut`, explicit using them in the table is strange, and may break the backward compatibility.

We can just remove it. Currenly, we don't generate memonics for segment load/store in both Golang/Rust. Yes, it is OK for Chisel and other HDL targets, but it is necessary for software porting.

sequencer:
I see, adding seg load store instruction explicitly look good to me. If this is merged, we can follow it.

wangpc-pp:
Ping.

wangpc-pp:
Ping.

aswaterman:
I want to see the Spike PR before merging this.  (Spike should be essentially unchanged; we shouldn't end up having to make dozens of copies of the existing instructions just because of this riscv-opcodes PR.)

PR #272, opened 2024-08-01
Generate 0xFFFFFFFF instead of 0 for rs1 field in make_go

wangpc-pp:
The default rs1 is 0, but this may make encoding process hard to distinguish the real rs1 of an instruction.

We set it to 0xFFFFFFFF instead.

### DIFF
--- a/parse.py
+++ b/parse.py
@@ -966,9 +966,10 @@ func encode(a obj.As) *inst {
     instr_str = ''
     for i in instr_dict:
         enc_match = int(instr_dict[i]['match'],0)
+        enc_encoding = instr_dict[i]['encoding']
         opcode = (enc_match >> 0) & ((1<<7)-1)
         funct3 = (enc_match >> 12) & ((1<<3)-1)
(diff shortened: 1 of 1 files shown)


wangpc-pp:
cc @4a6f656c

aswaterman:
This is hacky.  What about other register specifiers?  What about cases where only part of the register specifier is a don't-care?  (e.g. the forthcoming Zpa extension uses the LSB of some register specifiers as an opcode bit, but the other 4 bits remain significant)

wangpc-pp:
> This is hacky. What about other register specifiers? What about cases where only part of the register specifier is a don't-care? (e.g. the forthcoming Zpa extension uses the LSB of some register specifiers as an opcode bit, but the other 4 bits remain significant)

Yes, you are right. This change is highly related to the implementation, I will hold it util we have decided how to implement it in Golang.

4a6f656c:
We almost certainly do not want to do this - each instruction has a known encoding type and we only make use of the rs1 value provided in the instruction encoding, if the encoding type specifies that we should do so.

Issue #263, opened 2024-07-18
Include nf field in instruction name


moscickimilosz:
The nf field is treated as a variable arg within the rv_v instruction encoding file.

As I am interested in having all rvv instructions present in the YAML output, I was wondering whether a PR that expands this would get accepted? I believe this would align better with other extensions present in the dict e.g. Zimop and Zcmop.

I can see 3 possible solutions:
1. Expand rv_v file to include all of the instructions
2. Expand rv_v file to include all of the instructions, with the new instructions encoded as $pseudo_op (following what is done in rv_zimop and rv_zcmop)
3. Edit parse.py to expand dictionary only for the YAML output - avoid altering other outputs.

aswaterman:
Whatever we end up doing here, for backwards compatibility we need to make sure that the existing definitions that end up in `encoding.out.h`, `inst.chisel`, etc. are retained.

moscickimilosz:
> Whatever we end up doing here, for backwards compatibility we need to make sure that the existing definitions that end up in `encoding.out.h`, `inst.chisel`, etc. are retained.

Hi @aswaterman, 
Could you please take a look at the proposed solution in this [PR](https://github.com/riscv/riscv-opcodes/pull/274)?

Thanks

PR #274, opened 2024-08-01
Expand nf fields in instr_dict.yaml

moscickimilosz:
parse.py has been edited to expand nf field for segmented vector load store instructions. For backwards compatibility this only applies to instr_dict.yaml.

### DIFF
--- a/parse.py
+++ b/parse.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 from constants import *
+import copy
 import re
 import glob
 import os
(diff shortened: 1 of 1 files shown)


moscickimilosz:
This is a backwards compatible solution to a problem discussed in https://github.com/riscv/riscv-opcodes/issues/263.

Ideally my preference would be to expand this into other formats and break the backwards compatibility by hiding this functionality under a (disabled by default) command line flag - would probably require refactoring parse.py to use argparse first.

aswaterman (approved):

PR #275, opened 2024-08-21
Zicntr was inside Zicsr file

AFOliveira:
Even though, Zicntr depends on Zicsr, separating their instructions in different files allows for easier parsing when use cases of this repo actually need extension names. Aditionally, the  ".H" only exists when XLEN == 32, so to follow the repo normal guidelines it should sit on it's own file.


aswaterman (approved):

aswaterman:
@AFOliveira Following this logic, shouldn't you also move all of the FP-related pseudo-ops ones into `rv_f`?  You can do so as a follow-on PR if you'd like.

AFOliveira:
> @AFOliveira Following this logic, shouldn't you also move all of the FP-related pseudo-ops ones into `rv_f`?  You can do so as a follow-on PR if you'd like.

Definitely. I will move them and PR as soon as I can. Thank you!

PR #276, opened 2024-08-22
Move F extension CSRs to the rv_f

AFOliveira:
As mentioned in #275, this PR moves "F" extension pseudo-ops that were in the rv_zicsr to rv_f


AFOliveira:
On a slightly different topic, I also found that several pseudo-instructions from the ASM manual are not described in this repo (like mv, nop, bleu, bgtu), is there any reason for this or shall I go ahead and commit the ones I can find?

aswaterman:
I confess I don't know what the discipline for including or excluding pseudo-ops has been.  For my purposes, adding MV, NOP, etc. would not add any value, but I can't speak for others.

aswaterman (approved):

AFOliveira:
If this repo had all the instructions and pseudo-instructions in the ISA, I think it would allow for a range of tools to use them as a DB. 
I don't know if you've checked out this PR I made  riscv/riscv-isa-manual/pull/1603, but having all the instructions and pseudo instructions here in riscv-opcodes would make it possible to generate a full list in .adoc and then pdf.  Whatever the use case for that might end up on being I think that it illustrates one thing in which this addition could help.

PR #278, opened 2024-08-27
Added the rest of 1:1 PseudoInstructions

AFOliveira:
Complementing #277, there were a few 1:1 pseudo instructions that were missing.
Below is the list with the link to the asm manual:

# I 
[j](https://github.com/riscv-non-isa/riscv-asm-manual/blob/f8bcdded42ac9108e0bf7cf0789dbe306ec329e2/riscv-asm.md?plain=1#L961)
[jal ](https://github.com/riscv-non-isa/riscv-asm-manual/blob/f8bcdded42ac9108e0bf7cf0789dbe306ec329e2/riscv-asm.md?plain=1#L962)
[jr](https://github.com/riscv-non-isa/riscv-asm-manual/blob/f8bcdded42ac9108e0bf7cf0789dbe306ec329e2/riscv-asm.md?plain=1#L963)
[jalr ](https://github.com/riscv-non-isa/riscv-asm-manual/blob/f8bcdded42ac9108e0bf7cf0789dbe306ec329e2/riscv-asm.md?plain=1#L964)

# Zicsr
[csrr](https://github.com/foss-for-synopsys-dwc-arc-processors/riscv-opcodes/blob/436c27e1607f56e23919179218d212ee1b6716d8/rv_zicsr#L9) 
[csrw](https://github.com/foss-for-synopsys-dwc-arc-processors/riscv-opcodes/blob/436c27e1607f56e23919179218d212ee1b6716d8/rv_zicsr#L10)
[csrs](https://github.com/foss-for-synopsys-dwc-arc-processors/riscv-opcodes/blob/436c27e1607f56e23919179218d212ee1b6716d8/rv_zicsr#L11)
[csrc](https://github.com/foss-for-synopsys-dwc-arc-processors/riscv-opcodes/blob/436c27e1607f56e23919179218d212ee1b6716d8/rv_zicsr#L12)
[csrwi](https://github.com/foss-for-synopsys-dwc-arc-processors/riscv-opcodes/blob/436c27e1607f56e23919179218d212ee1b6716d8/rv_zicsr#L13)
[csrsi](https://github.com/foss-for-synopsys-dwc-arc-processors/riscv-opcodes/blob/436c27e1607f56e23919179218d212ee1b6716d8/rv_zicsr#L14)
[csrci](https://github.com/foss-for-synopsys-dwc-arc-processors/riscv-opcodes/blob/436c27e1607f56e23919179218d212ee1b6716d8/rv_zicsr#L15)

### DIFF
--- a/rv_i
+++ b/rv_i
@@ -70,3 +70,7 @@ $pseudo_op rv_i::sltu   snez rd rs2 31..25=0 19..15=0x0 14..12=3 6..2=0x0C 1..0=
 $pseudo_op rv_i::slt    sltz rd rs1 31..25=0 24..20=0x0 14..12=2 6..2=0x0C 1..0=3
 $pseudo_op rv_i::slt    sgtz rd rs2 31..25=0 19..15=0x0 14..12=2 6..2=0x0C 1..0=3
 
+$pseudo_op rv_i::jalr   jalr rs1    31..20=0 14..12=0 11..7=0x01 6..2=0x19 1..0=3
+$pseudo_op rv_i::jalr   jr   rs1    31..20=0 14..12=0 11..7=0x0  6..2=0x19 1..0=3
+$pseudo_op rv_i::jal    jal  jimm20                   11..7=0x01 6..2=0x1b 1..0=3
+$pseudo_op rv_i::jal    j    jimm20                   11..7=0x0  6..2=0x1b 1..0=3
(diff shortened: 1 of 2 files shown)


aswaterman (approved):
I was worried that pseduo-ops with the same name as real instructions might cause problems somewhere, but so far, everything seems OK.  We can always roll back the addition of `jal` and `jalr` pseudoinstructions in the future if need be.

AFOliveira:
I think same name pseudo ops won't be the problem. However there are some pseudo-instructions that are not possible to reference with today's structure because they are caught as errors on parse.py. I will try to do a minimalistic change that will allow for further pseudo-ops definition and submit a PR with it.

aswaterman:
I think it would be much cleaner to have the script detect this case automatically (by observing the name collision) and do whatever is necessary.  We don't want the extra boilerplate `pseudo_`; the `$pseudo_op` directive should be smart enough to do whatever is necessary.

AFOliveira:
Ok, then, I will do that and reference this PR then. Thank you.

aswaterman:
Thank you for your help, @AFOliveira!

PR #290, opened 2024-10-02
Add c.sext.w and zext.w

AFOliveira:
I'm now looking on to binutils and the ISA and trying to fix gaps between all the sources. 

[c.sext.w ](https://github.com/bminor/binutils-gdb/blob/2754d75a11556577fcddf23718bfbff7d0489a3f/opcodes/riscv-opc.c#L2179)

[zext.w](https://github.com/bminor/binutils-gdb/blob/2754d75a11556577fcddf23718bfbff7d0489a3f/opcodes/riscv-opc.c#L1274)

I also found that this repo doesn't have the any instruction with acquire and release, like "[amoswap.b.rl](https://github.com/bminor/binutils-gdb/blob/2754d75a11556577fcddf23718bfbff7d0489a3f/opcodes/riscv-opc.c#L714)", is there a reason for these, or can i also commit them?


rv64_zcb
```diff
@@ -1 +1,3 @@
 c.zext.w  rd_rs1_p 1..0=1 15..13=4 12..10=7 6..5=3 4..2=4
+
+$pseudo_op rv64_c::c.addiw c.sext.w  rd_rs1_n0 15..13=1 12=0 6..2=0 1..0=1
```
aswaterman, line 3: `c.addiw` is part of the C extension, not Zcb, so this should be moved to `rv64_c`.

aswaterman: It looks like Binutils picks up `sext.w` as a pseudoinstruction for `c.addiw` when the C extension is enabled: https://github.com/bminor/binutils-gdb/blob/2754d75a11556577fcddf23718bfbff7d0489a3f/opcodes/riscv-opc.c#L580

But it only picks up `c.sext.w` as a pseudoinstruction for `c.addiw` when Zcb is enabled.

This all seems pretty illogical to me, but I guess we are stuck with the binutils behavior at this point for compatibility reasons, right @kito-cheng?

AFOliveira: Although it seems ilogical, my guess is binutils is doing that because the pseudo-instruction is only mentioned as a note on Zcb, maybe if the ISA added that, we could change all the others.

AFOliveira: @aswaterman Just a ping. Do we want to keep this on hold? 
Technically, the PR is according to the rest of the sources. if anything changes in the meantime, I'll be happy to address those changes and PR any update.

aswaterman (approved):

PR #291, opened 2024-10-04
Put main code in a function

Timmmm:
This is a small change that has a couple of advantages:

1. It lets other Python code import this module and run `main()`.
2. When this code is converted to a modern pyproject-based setup it will be required in order to expose this script.

### DIFF
--- a/parse.py
+++ b/parse.py
@@ -118,7 +118,7 @@ def process_enc_line(line, ext):
                 existing_arg, new_arg = parts
                 if existing_arg in arg_lut:
                     arg_lut[a] = arg_lut[existing_arg]
-                
+
                 else:
                     logging.error(f' Found field {existing_arg} in variable {a} in instruction {name} whose mapping in arg_lut does not exist')
                     raise SystemExit(1)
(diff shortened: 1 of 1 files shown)


aswaterman:
@IIITM-Jay please review.

parse.py
```diff
@@ -1046,7 +1046,7 @@ def signed(value, width):
     return value - (1<<width)
 
 
-if __name__ == "__main__":
+def main():
```
IIITM-Jay, line 1049: @aswaterman yes, we can do this change as I have already did that in Refactored `parse.py` file in my PR ( a longer one). So yeah, this could be accepted.

aswaterman: Let's close this PR and handle this as part of merging your PR.

Timmmm: @IIITM-Jay which PR are you referring to? I can't find one that does this.

IIITM-Jay: @Timmmm  I have revised the PR #283 with mentioning in comments, about only to consider that PR with pure intention to refactor and optimize the parser logic with modular approach to enhance maintainability and readability keeping the best practices and standards.

Will include this thing as I proceed further refactoring other scripts which are dedicated to generate the corresponding artifacts such `encoding.out.h`, `instr-table.tex` etc. 
I have in mind to include the changes w.r.t `main` method inclusion.

PR #292, opened 2024-10-04
Add basic pre-commit config, apply fixes and run it in CI

Timmmm:
Add a basic pre-commit config to format Python files using Black, and run it in CI. The second commit in this MR contains the actual formatting fixes.

I will fix the Pylint failures and add types in future MRs so that they can be enabled.


aswaterman:
This is a misleading PR title; it suggests no changes to the actual code. I don’t have a problem with the PR, of course, but please do try to employ accurate PR titles.

@IIITM-Jay please review.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+
+  - repo: https://github.com/pycqa/isort
+    rev: 5.13.2
+    hooks:
+      - id: isort
+        # Tell isort to output in a Black-compatible style, otherwise you
+        # can get into a situation where isort changes a file and Black changes
+        # it back, so pre-commit never passes.
```
IIITM-Jay, line 16: I think use of pronouns (such as  "you" here) is not best practice to write comments IMO and it also reflects sometime informal and directive. We can change the above comment something to like _(Just a suggestion)_:

```
# Configure isort to use Black's import style to prevent formatting conflicts.
# As isort and Black may repeatedly reformat the same code, causing pre-commit to fail.
```

.github/workflows/python-app.yml
```diff
@@ -15,10 +15,14 @@ jobs:
...
+    - name: Run pre-commit
+      run: |
+        python3 -m pip install pre-commit
+        pre-commit run -A
     - name: Install PyYAML
       run: |
-         pip3 install -r requirements.txt
-         pip3 install coverage
+        pip3 install -r requirements.txt
+        pip3 install coverage
```
IIITM-Jay, line 25: @aswaterman In my view, and also thinks of a best practice and approach here, that reordering of the steps: Installing Dependencies and Running Pre-Commit -  to be interchanged.

What I meant to say that, installing dependencies before running pre-commit hooks, will ensure that any packages required by the hooks are available, preventing potential failures during the pre-commit phase.

I think the sequence to be like:
```
# First Installation then pre commit

name: Install PyYAML
      run: |
         pip3 install -r requirements.txt
         pip3 install coverage

# Now run pre commit
 name: Run pre-commit
      run: |
        python3 -m pip install pre-commit

```

.github/workflows/python-app.yml
```diff
@@ -15,10 +15,14 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
+    - name: Run pre-commit
+      run: |
+        python3 -m pip install pre-commit
+        pre-commit run -A
```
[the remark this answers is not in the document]
Timmmm: I totally agree; I'll fix it.

.github/workflows/python-app.yml
Timmmm: Yep that makes sense. I'll change it.

.github/workflows/python-app.yml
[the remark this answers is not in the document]
Timmmm: Done

.pre-commit-config.yaml
Timmmm: Done

.github/workflows/python-app.yml
Timmmm: Done

aswaterman (approved):
Thanks to the both of you.

aswaterman:
@Timmmm @IIITM-Jay After merging, a post-merge check is failing, probably because of a formatting error that has been inserted since this PR was created.  Can either of you look into it and file a PR to get things passing again?

https://github.com/riscv/riscv-opcodes/actions/runs/11225349353/job/31203877551

Timmmm:
Ah probably a race with another PR. I'll fix it!

aswaterman:
Thank you!

Timmmm:
#296 should fix it.

PR #293, opened 2024-10-04
Add static type hints

Timmmm:
This makes the code easier to understand and navigate, and also detected a few of bugs:
    
1. Missing brackets on e.upper. (Fixed)
2. Not strictly related to types, but a lot of the regexes were not raw strings and therefore contained invalid escape sequences. Python prints a warning about these in recent versions. (Fixed)
3. Expression in `process_pseudo_instructions()` that is always false. (Not fixed)
4. Missing definition of `log_and_exit()`. (Fixed)
    
This is validated via pre-commit in CI.


Timmmm:
Before I rebase this (looks like it will be quite a big task unfortunately), @aswaterman can I get a nod that this would be acceptable so I don't waste my time?

aswaterman:
@Timmmm Yes, I agree static type hints are an improvement.  Sorry the refactoring will be painful.

I might gently suggest that you make the CI change to verify the static type hints as part of the same PR, so that we don't regress, but since you are the one doing the work, I'm not in a position to make any demands.

Timmmm:
Great, thanks!

> I might gently suggest that you make the CI change to verify the static type hints as part of the same PR, so that we don't regress, but since you are the one doing the work, I'm not in a position to make any demands.

Sounds like a good idea; I'll do that.

Timmmm:
Ok I have rebased, added it to CI/precommit and updated the commit & MR messages.

It is built on top of the commits from #306 and #304 since they made it easier and they are easy wins.

This found an additional issue (the missing `log_and_exit()` function) which I have fixed. The code around `ext_name` (now in `process_pseudo_instruction()`) has been changed since I started but it's still incorrect. I'm unsure what the intent is there so I left it as-is.

aswaterman (commented):
LGTM, except for the one nit that this appears to incorporate the changes being discussed in #306.  Can you either factor that out (which I know might or might not be a pain) or wait until that PR is resolved?

Timmmm:
Also #304. They're in separate commits though so I was just going to wait until they were merged (assuming it doesn't take too long) and then rebase.

IIITM-Jay:
> except for the one nit that this appears to incorporate the changes being discussed in https://github.com/riscv/riscv-opcodes/pull/306.

@aswaterman I have made some changes and left some comments out there, if you agree, we can proceed accordingly.

README.md
```diff
@@ -126,9 +126,11 @@ of extensions are being processed such that the *base-instruction* is not includ
 
 The following artifacts can be generated using parse.py:
 
-- instr\_dict.yaml : This is file generated always by parse.py and contains the
-  entire main dictionary `instr\_dict` in YAML format. Note, in this yaml the
-  *dots* in an instruction are replaced with *underscores*
+- instr\_dict.json : This is file generated always by parse.py and contains the
```
IIITM-Jay, line 129: @Timmmm I think the above can be replaced by _(just some grammatical improvements)_

_"This file is always generated by `parse.py` and contains the.... "_

README.md
```diff
@@ -220,6 +214,6 @@ DEBUG::      Processing line: bne     bimm12hi rs1 rs2 bimm12lo 14..12=1 6..2=0x
 ## How do I find where an instruction is defined?
 
 You can use `grep "^\s*<instr-name>" rv* unratified/rv*` OR run `make` and open
-`instr_dict.yaml` and search of the instruction you are looking for. Within that
+`instr_dict.json` and search of the instruction you are looking for. Within that
```
IIITM-Jay, line 217: @Timmmm I think the above can be rewritten like:

_" `instr_dict.json` and **search for** the instruction you are looking for. Within that ... "_

README.md
Timmmm: Ooo good spot; I'll fix this and the other one in #304.

IIITM-Jay: @aswaterman I think @Timmmm has pushed the above changes in #304

README.md
IIITM-Jay: @aswaterman I think these are also being pushed #304

aswaterman (approved):

PR #294, opened 2024-10-04
Reorganise Python code using pyproject.toml

Timmmm:
Add pyproject.toml (the modern alternative to requirements.txt), making this a proper Python package that can be installed via pip and potentially uploaded to PyPI.

This required moving all of the data files into the package, and reading them using importlib. This gives a much nicer file structure. It does require Python 3.7 but that is over 6 years old and easy to install on old Linux distros like RHEL 8.

Note: This is a draft; I haven't quite finished the bit that actually reads the opcode files, but I wanted to get feedback to see if you're ok with this approach first.

Especially the bit about creating a venv. Python's native tooling is kind of terrible so this is not easy to avoid. If you install [rye](https://rye.astral.sh/) then it becomes a bit better; you can run `rye run riscv_opcodes ...` and it will take care of the venv automatically. Also end-users won't see the pain - if they do `pip3 install riscv_opcodes` then they will just be able to run `riscv_opcodes ...`.


aswaterman:
Solution to a non-problem IMO.

Timmmm:
We use riscv-opcodes in our Python flow and it's extremely awkward to use as-is. It would significantly easier if we could just install it as a package.

aswaterman:
Yeah, sure, I'll reconsider if we do it in a way that doesn't reorganize the directory structure of the extensions files themselves.

Timmmm:
I updated the PR so they are in the same layout - the `extensions` directory is moved but otherwise unmodified. Is that acceptable?

Timmmm:
Will do.

PR #268, opened 2024-07-22
Refactor and Optimization:: Argument Table Formation:: Constants.py

IIITM-Jay:
### About
This PR deals with refactoring and optimization towards Argument Lookup Table `(arg_lut)` of `constants.py`. It focuses on optimizing the code with fewer lines keeping the objective same to provide better readability and concise comprehension.

### Added/ Modified Changes

1. **Changed the formation of `arg_lut = {}`:** Used dictionary comprehension to form the required `arg_lut` 
2. **Read the CSV file and processes each row** to convert the necessary values and build the dictionary.
3. **Added appropriate comments**
4. **Avoided explicit iteration and assignment of values within a loop**
5. **Removed redundancy:** prior declaration of a dictionary  

### Approach/ Technique/ Principles Used

1. **Dictionary Comprehension:** 
   a. Offers syntax with lesser lines of codes: [Dict Comprehensions](https://peps.python.org/pep-0274/) , [Reference taken ](https://python-reference.readthedocs.io/en/latest/docs/comprehensions/dict_comprehension.html)
          `  {expression(variable): expression(variable) for variable, variable in input_set [predicate][, …]}`
   b. Does exactly the same job without having prior declaration of the dictionary variable

PR #296, opened 2024-10-08
Reapply pre-commit autoformat

Timmmm:
This little bit of code from #268 was not formatted in #292 because it was not merged yet. Basically it was a semantic conflict between the PRs, but not a text conflict, so Git can't detect it.

This can be prevented using merge queues, but it probably isn't necessary on a repo of this size.

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -43,7 +43,6 @@ imported_regex = re.compile(
 )
 
 
-
 def read_csv(filename):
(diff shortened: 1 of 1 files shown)


aswaterman:
deferring to @IIITM-Jay

aswaterman (approved):

PR #299, opened 2024-10-10
[enhc] support for multiple Python versions

Myrausman:
Refactored GitHub Actions workflow to support multiple Python versions


.github/workflows/tests.yml
```diff
@@ -9,19 +9,28 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.8', '3.10', '3.12']
```
IIITM-Jay, line 14: Similarly, here based on previous comment for maintaining consistency across files.

Myrausman:
Thank you for your feedback, @IIITM-Jay. I'll certainly delve deeper into the potential implications of our existing code on older Python versions. I'm encountering a minor challenge with my GitHub Action workflow. While attempting to utilize Python 3.6, I'm receiving the error `Version 3.6 was not found in the local cache.`

Could you please provide some guidance on how to resolve this issue?

IIITM-Jay:
Oh! I see, Sorry for the delayed response!

@aswaterman and @Myrausman , The error that you are facing with this PR currently is that `Python 3.6` is no longer available in the GitHub Actions environment. Python 3.6 reached its   [end-of-life (EOL) on December 23, 2021](https://devguide.python.org/versions/#versions) and is no longer supported or maintained. The link will also guide the EOL of other versions in concern.

But, if we really want to test against `Python 3.6`, we have few options for the same:

1. We can use **Docker Container** that has Python 3.6 installed. It will act as an Isolated environment without needing a self-hosted runner. Will come to this self-hosted runner, in next point. But it may come up with code complexities.
2. We can have our own **Self-Hosted Runner** by setting up a self-hosted GitHub Actions with Python 3.6 installed. This allows  to maintain control over the Python versions available. But again, requires maintenance and management of the runner.
3. **Legacy `actions/setup-python` Versions:** I am not sure with this, but I think, older versions of `actions/setup-python` might still support Python 3.6. Again, I can't suggest this one here due to potential security vulnerabilities and lack of support.  For the sake of simplicity to make my point clear, I am thinking of this :
```
- name: Set up Python 3.6
  uses: actions/setup-python@v1
  with:
    python-version: '3.6'
```

IIITM-Jay:
@aswaterman , few more suggestions:

1. We can use updated GitHub Actions syntax to use the latest versions of GitHub Actions for better features and security.I meant about `actions/checkout` and `actions/setup-python`.
2. We can specify python dependencies efficiently by combining the pip install commands to reduce the number of steps and improve caching.
3. We can speed up the workflows by utilizing cache.

Thinking to do above related things in a separate PR elaborating each steps and the reasons why we are adopting those

sequencer:
I wonder should the python parser separates from this repo as a standalone project? Originally, it was possible. But recently the trend of it seems to grow into some big project that even needs multiple python versions support. It’s kinda unbelievable for such a simple project. 

IMHO, the change to parser logic is always unrelated to the opcodes itself, and seldom does the opcode changes(only add new instructions and move ops from unratified).

I think we maybe can formally define the syntax of opcodes and move the parser logic to another repo if the python parser is too mutable.

aswaterman:
I'm OK with ditching support for Python 3.6, so we can revise this PR to only test against 3.8 and above if that's what we decide is best.

And I support any sensible improvements to improve the extent of our CI or speed it up.

.github/workflows/tests.yml
```diff
@@ -9,19 +9,28 @@ on:
...
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install dependencies
       run: |
-         pip3 install -r requirements.txt
-         pip3 install coverage
+         pip install -r requirements.txt
+         pip install coverage
+
     - name: Test error outputs
```
IIITM-Jay, line 29: @Myrausman we should keep `pip3` as it will intend to `python 3.x` packages and it is specifically associated with `Python 3.x` versions. 
While the only `pip` refers to the default package manager for `Python 2.x`. Additionally, on systems where `Python 2` is installed, `pip` installs packages for `Python 2`. So, in order to avoid conflicts between `Python 2.x` and `Python 3.x`, we should keep as `pip3`.

.github/workflows/python-app.yml
```diff
@@ -9,25 +9,36 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.6', '3.8', '3.10', '3.12']
```
IIITM-Jay, line 14: @Myrausman we can have all the current supported versions of `Python 3.x` and not the ones which are no longer supported.

IIITM-Jay (changes requested):
@Myrausman I have left some comments to revise this PR.

@aswaterman I think it would be good to showcase with the help of this CI setup that this repository is in-line with all the current supported versions of `Python 3.x`. I am taking the reference from the [Current Status of Python versions](https://devguide.python.org/versions/#versions) , to revise this PR.

IIITM-Jay:
@Myrausman as per @rpsene, we would be doing these for `python versions` **3.9 to 3.12** as they all are supportive according to the [Current Status of Python versions](https://devguide.python.org/versions/#versions)

IIITM-Jay (approved):
Approving it as it shows merging is blocked until the requested changes are being addressed (as I requested for those changes). 
@aswaterman needed your suggestions to go ahead. What do you think for the support of the other `python versions` mentioned in this PR.

aswaterman (approved):

PR #300, opened 2024-10-14
ci: use riscv-builders

mengzhuo:
To runs with riscv-builders, we need to install Github App
https://github.com/apps/riscv-builders

image

p.s It takes 9m to complete the CI, although it's ~30s with x86

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -8,20 +8,18 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: riscv-builders
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python 3.8
-      uses: actions/setup-python@v2
-      with:
-        python-version: 3.8
     - name: Install PyYAML
       run: |
-        pip3 install -r requirements.txt
-        pip3 install coverage
+        apt install -yq python3-pip
+        pip config set global.break-system-packages true
+        pip install -r requirements.txt
+        pip install coverage
     - name: Run pre-commit
       run: |
-        python3 -m pip install pre-commit
+        pip install pre-commit
         pre-commit run --all-files
     - name: Generate
       run: coverage run ./parse.py -c -chisel -sverilog -rust -latex -spinalhdl -go "rv*" "unratified/rv*"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,17 +8,15 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


IIITM-Jay:
The reason checks are still running is that it is searching for the runner `riscv-builders` and it will search for indefinite time and that is forever. It will not be able to get the `riscv-builder` runner at all as it is acting as a **_self-hosted runner_** for this repository and it is trying to get that from GitHub Actions Environment. 

@aswaterman for this CI to get run, we have to configure and setup the same in a self-hosted runners. Currently, we are using the GitHub hosted runners which provides the OS(s) - Windows, Linux and Mac. Out of the provided OS as stated, we are running the CI on the ubuntu system of the latest version.

I am curious to know @aswaterman if you could throw some light on `riscv-builders`. As why it is beneficial to run and test this repository on `riscv-builders` and does it really necessary to run the builds on the same. Are the RISC-V related repositories suppose to be run on `riscv-builders` instead of the defualt OS provided by GitHub. 

I am in doubt as the hosted GitHub pages for `riscv-builders` says : _**"This project is very early alpha stage and NO WARRENTY."**_
https://riscv.builders/#sla

mengzhuo:
@IIITM-Jay Thanks for the reply, what is requirement that we need for install a Github app? I might setup a topic on RISE TSC meeting on that. As for SLA, I can buy a sever rack in data center to improve SLA.

@aswaterman The main reason is I've got PR rejected by some open source projects by "what kind of CI that runs RISCV?" https://github.com/goreleaser/goreleaser/pull/5196#issuecomment-2408587609, https://github.com/open-telemetry/opentelemetry-go/pull/5420#issuecomment-2135414810, I think it's time to leading by example or ["dogfooding"](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) now.

IIITM-Jay:
> This software is meant to run on any POSIX system regardless of ISA, so we should use whatever CI system is cheapest/easiest to use/etc.

@aswaterman Yes, I completely agree with this. We should focus on the cheapest, and the easiest one to maintain on a longer run thereby using the best and standard practices. Additionally, we should also focus on minimizing the time it takes to run the CI builds to completion as higher times are not the best time acceptable for the CI to take. Here, as above, it took **_9 minutes_** for the builds to complete _(as stated in the screenshot)_ which is not quite good enough.

aswaterman:
Yeah, we're on the same page, @IIITM-Jay.

@mengzhuo I appreciate what you're trying to do, but this is not the right place to do it.  The focus should be on software that actually runs on RISC-V processors.  For the time being, this is not that software.

PR #304, opened 2024-10-27
Generate JSON instead of YAML

Timmmm:
Previously this generate a `instr_dict.yaml` file, however there is no need to use YAML here. The only advantages of YAML over JSON are that it is (debatably) easier for humans to write, making it suitable as human-edited input files. This is a machine-generated output file so JSON is better.

The main advantages are:

1. It removes the dependency on PyYAML, which is the only external dependency of this project. Python external dependencies are quite a pain - PyYAML is a particularly troublesome one - and having no dependencies is very convenient (no need for venv etc.).
2. It means consumers of the file don't need to depend on PyYAML.

Note this could have been done in an 100% backwards compatible way by keeping the file name unchanged (since JSON is valid YAML), however I figured there probably aren't that many users of this file, and since the only thing they need to change is the filename it's probably better to minimise confusion by renaming it. It also makes it clear that it is guaranteed to be JSON.


IIITM-Jay:
Just keeping my thoughts here @aswaterman regarding the above mentioned points in the PR description



>  This is a machine-generated output file so JSON is better.

@Timmmm Even though this is machine-generated, YAML’s indentation-based structure makes large, deeply nested data easier to read at a glance than JSON’s curly-brace heavy syntax. It is easier to review and debug faster comparable to JSON structure.

For an example,  YAML
```
instructions:
  - name: ADD
    opcode: 0x01
    description: "Adds two numbers"
```

and JSON,
```
{
  "instructions": [
    {
      "name": "ADD",
      "opcode": "0x01",
      "description": "Adds two numbers"
    }
  ]
}
```

- Although the current output is machine-generated, requirements might change in the future. YAML offers more flexibility to grow organically with minimal changes.
  - For example, if configuration overrides, custom templates, or metadata annotations become necessary in the future, YAML will handle them more elegantly.
  - Choosing YAML ensures the project is prepared for evolving needs without requiring a switch back from JSON later.
 
> It removes the dependency on PyYAML, which is the only external dependency of this project. Python external dependencies are quite a pain

- While the argument against PyYAML is reasonable (avoiding external dependencies), YAML parsing is trivial in modern Python. Additionally, Python’s standard library has added support for TOML (another human-friendly format), showing a trend toward readable config formats. In contrast, JSON is less suited for these roles, _potentially making YAML a better long-term choice._

> Note this could have been done in an 100% backwards compatible way by keeping the file name unchanged (since JSON is valid YAML)

Absolutely correct, YAML files are a superset of JSON. If a consumer prefers JSON, they can parse the YAML as JSON directly, without needing PyYAML. Here I could say that:  **_Keeping YAML does not limit interoperability—it just offers more flexibility._**

parse.py
```diff
@@ -46,8 +45,10 @@
 
     instr_dict = create_inst_dict(extensions, include_pseudo)
 
-    with open("instr_dict.yaml", "w") as outfile:
-        yaml.dump(add_segmented_vls_insn(instr_dict), outfile, default_flow_style=False)
+    with open("instr_dict.json", "w") as outfile:
+        json.dump(
+            add_segmented_vls_insn(instr_dict), outfile, ensure_ascii=False, indent=2
+        )
```
IIITM-Jay, line 51: @Timmmm Also, I could see this: any specific reason to mention `ensure_ascii` as `False` and why we need to explicitly state that parameter.

Timmmm: It means it will use UTF-8 encoding for the JSON, however I don't think we'll have any non-ASCII anyway so I'll remove it for simplicity.

Timmmm:
> It is easier to review and debug faster comparable to JSON structure. 

Well, I could debate the "improved" readability of YAML. Would you have guessed from the YAML that `0x01` is a string? I would have to look that up, whereas it's very clear in the JSON.

But in any case I don't think you can argue that the difference in readability is *significant*, whereas the benefit of not relying on PyYAML and not requiring consumers to deal with YAML are undeniably large.

parse.py
```diff
@@ -44,8 +43,8 @@
 
     instr_dict = create_inst_dict(extensions, include_pseudo)
 
-    with open("instr_dict.yaml", "w") as outfile:
-        yaml.dump(add_segmented_vls_insn(instr_dict), outfile, default_flow_style=False)
+    with open("instr_dict.json", "w") as outfile:
+        json.dump(add_segmented_vls_insn(instr_dict), outfile, indent=2)
```
IIITM-Jay, line 47: @Timmmm and @aswaterman , just thinking, whether it would be a better idea to wrap the file operations in a `try-except` block for handling file writing errors _(e.g., permission issues, disk space problems)_, like below, for an example:

```
try:
    with open("instr_dict.json", "w") as outfile:
        json.dump(add_segmented_vls_insn(instr_dict), outfile, indent=2)
except IOError as e:
    print(f"Error writing to file: {e}")
```

Timmmm: Yeah I agree; reading Python backtraces isn't fun. I think we should do it in another PR though probably.

Timmmm:
I fixed the two grammar mistakes in the README you noticed.

parse.py
aswaterman: SGTM too.

Issue #289, opened 2024-10-01
Missing B Extensions in `encoding.h`


IIITM-Jay:
@aswaterman As I was going through the `encoding.h` file, I observed that B-extensions are not present. It may be probably because the files with  prefixes `rv32_zb*`  (e.g., `rv32_zbb`, `rv32_zbs`) are having `pseudo_op` and it is being excluded.

In the parsing logic, the flow of `parse.py` too ignores `pseudo_op` .
Is these extensions need not be included in `encoding.h` ?

aswaterman:
The B extension is just Zba + Zbb + Zbs, so is it really missing?  The contents of `rv_zba` etc. appear to be complete.

Although it would be desirable to include pseudoinstructions in `encoding.h`, we need to contend with name collisions.  For example, `jal` is both a real instruction and a pseudoinstruction.  (I don't think we have any cases where we have a 3-way collision, so maybe just prepending the instruction with `PSEUDO_` would suffice...?  e.g. `MATCH_PSEUDO_JAL`, `MASK_PSEUDO_JAL`.  Just a thought; not sure this is the optimal solution.)

AFOliveira:
>  As I was going through the encoding.h file, I observed that B-extensions are not present. It may be probably because the files with prefixes rv32_zb* (e.g., rv32_zbb, rv32_zbs) are having pseudo_op and it is being excluded.

I just wanted to point out that this is not true, if you check per example "sh1add", that is present even without pseudo-ops. Additionally, if you do the default make, you will have all instructions(even the pseudo-ops). 

> In the parsing logic, the flow of parse.py too ignores pseudo_op .

The reason parse.py ignores the pseudo-ops when you don't run the automated make is that only including "go" is setting them on(see https://github.com/riscv/riscv-opcodes/blob/85da0d6a2e5b349f19e57b122abab74073aa44c6/parse.py#L1059C1-L1061C1). 
I don't know why this was built this way, but if anyone needs to include those, you can simply activate the pseudo_ops on a downstream fork and you will have all.

IIITM-Jay:
> The B extension is just Zba + Zbb + Zbs, so is it really missing? The contents of `rv_zba` etc. appear to be complete.

@aswaterman just thinking, if it would be beneficial and adds any value to the existing contents of  `zbb` , when added additional instructions such as `rev` (bit reversal) and `shfl/unshfl` (shuffle/unshuffle).

[...]

Issue #286, opened 2024-09-26
Unexpected jalpseudo/jalrpseudo instructions


4a6f656c:
Commit 13be4eee8ad2402d41699cf733a2fc19d12593ba added additional pseudo instructions, including these two lines in `rv_i`:

```
$pseudo_op rv_i::jalr   jalr rs1    31..20=0 14..12=0 11..7=0x01 6..2=0x19 1..0=3
$pseudo_op rv_i::jal    jal  jimm20                   11..7=0x01 6..2=0x1b 1..0=3
```

These result in `./parse.py -go` generating `AJALPSEUDO` and `AJALRPSEUDO` instructions, which are unexpected. It seems rather strange to alias an instruction to itself - the only difference between these and the original definitions appears to be that the register is hardcoded to `X1`. Have these been misnamed or added in error?

AFOliveira:
> These result in `./parse.py -go` generating `AJALPSEUDO` and `AJALRPSEUDO` instructions, which are unexpected. It seems rather strange to alias an instruction to itself - the only difference between these and the original definitions appears to be that the register is hardcoded to `X1`. Have these been misnamed or added in error? 

That's on purpose. If you check the ASM Manual (the exact links are on PR #278) that's how the pseudo-instructions are specified. I added the "pseudo" naming for those two variables so that the re-definition would not break any output already being used downstream. Moreover, I believe the reason for x1 being hardcoded there is that x1 actually is the Return Address specified on RISC-V ABI, but I guess @aswaterman is the best person to answer to that.

aswaterman:
> I believe the reason for x1 being hardcoded there is that x1 actually is the Return Address specified on RISC-V ABI

Right.

4a6f656c:
I understand why x1 is hardcoded here, however it seems to make little sense to provide an instruction called `AJALPSEUDO` which has this encoding. In reality, this is an assembly implementation issue whereby if no register is specfied for the `AJAL` instruction, it should default to x1.

I guess we'll just have to work around this by disabling the inclusion of pseudo-instructions for Go.

AFOliveira:
This repo being a source of information, I believe it should be as complete and possible and therefore hold the most information possible. However, I see how the changes may affect downstream application, but now that you're switching that I realized that maybe passing pseudo as a flag could work best. 

@IIITM-Jay Since you've been handling that refactor do you agree with this flag type of make? Technically, it would also solve #289

PR #305, opened 2024-10-27
Minor simplification of Python boolean

Timmmm:


parse.py
```diff
@@ -40,9 +40,7 @@
     extensions = [ext for ext in extensions if ext not in targets]
     print(f"Extensions selected : {extensions}")
 
-    include_pseudo = False
-    if "-pseudo" in sys.argv[1:]:
-        include_pseudo = True
+    include_pseudo = "-pseudo" in sys.argv[1:]
```
IIITM-Jay, line 43: I agree to the concise approach being followed here to reduce the number of lines of codes. Given that we should focus on optimizing the codes as much as possible, but at the same time, we should not forget its maintainability and readabaility.

Following, **_"Explicit is better than Implicit"_**: the core tenet of Python, the already existing codes explicitly communicates the default state for the inclusion of `pseudo instructions` as `(False)` and only sets it to `True `conditionally, making the flow of logic clearer. Readers can follow the intent more easily than inferring the meaning from a single line assignment. 

Additionally, I am tagging the discussions for reference regarding `pseudo instructions` inclusion for more information as to why we are using flags for explicit behaviour : #286 and #289

IIITM-Jay (commented):
@Timmmm I have left some comments as to why we are following the existing approach for better readability.

parse.py
Timmmm: I can't say I agree here. This isn't more explicit; it's just more verbose.

This is standard best practice. See [this blog for example](https://adamj.eu/tech/2020/01/17/simplify-your-ifs-that-return-booleans/#assignments).

In the long term it would be much better to use `argparse` of course, but this is a pretty straightforward simplification.

aswaterman (approved):
I agree with @Timmmm on this one.  Since it isn't an esoteric construction, the conciseness improves readability.

PR #306, opened 2024-10-27
Remove unused imports

Timmmm:
Remove unused Python import statements.

### DIFF
--- a/c_utils.py
+++ b/c_utils.py
@@ -2,7 +2,6 @@ import logging
 import os
 import pprint
 
-# from shared_utils import overlaps, overlap_allowed, extension_overlap_allowed, instruction_overlap_allowed, process_enc_line, same_base_isa, add_segmented_vls_insn, expand_nf_field
 from shared_utils import *
 
 pp = pprint.PrettyPrinter(indent=2)
--- a/c_utils.py
+++ b/c_utils.py
@@ -1,12 +1,6 @@
-import collections
-import glob
 import logging
 import os
 import pprint
-import re
-import sys
-
-import yaml
 
 # from shared_utils import overlaps, overlap_allowed, extension_overlap_allowed, instruction_overlap_allowed, process_enc_line, same_base_isa, add_segmented_vls_insn, expand_nf_field
 from shared_utils import *
--- a/go_utils.py
+++ b/go_utils.py
@@ -1,12 +1,6 @@
-import collections
-import glob
 import logging
-import os
 import pprint
-import re
 import sys
-
-import yaml
 
 # from shared_utils import overlaps, overlap_allowed, extension_overlap_allowed, instruction_overlap_allowed, process_enc_line, same_base_isa, add_segmented_vls_insn, expand_nf_field
 from shared_utils import *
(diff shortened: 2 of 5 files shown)


c_utils.py
```diff
@@ -1,12 +1,6 @@
...
-import glob
 import logging
 import os
 import pprint
-import re
-import sys
-
-import yaml
 
 # from shared_utils import overlaps, overlap_allowed, extension_overlap_allowed, instruction_overlap_allowed, process_enc_line, same_base_isa, add_segmented_vls_insn, expand_nf_field
```
IIITM-Jay, line 5: @Timmmm there are many related functions from `shared` module as frequently used across various scripts, listing them all will lead to reducing the readability of the code. The intention to use Wildcard imports was to declutter the import section, shifting the reader’s focus to the core logic rather than the long list of helper utilities. 

Moreover, it helps to keep the code more cleaner and concise, making the code easier to scan at a glance without a long list of imports at the top.
Hence, I guess these could be removed from commented out line.

go_utils.py
```diff
@@ -1,13 +1,7 @@
...
-import glob
 import logging
-import os
 import pprint
-import re
 import sys
 
-import yaml
-
 # from shared_utils import overlaps, overlap_allowed, extension_overlap_allowed, instruction_overlap_allowed, process_enc_line, same_base_isa, add_segmented_vls_insn, expand_nf_field
```
IIITM-Jay, line 5: Same as previous comment.

c_utils.py
Timmmm: Best practice is to avoid wildcard imports in general. The conciseness is not worth the downsides. See

* http://docs.quantifiedcode.com/python-anti-patterns/maintainability/from_module_import_all_used.html
* https://pybit.es/articles/why-you-should-avoid-import-in-python/
* https://stackoverflow.com/questions/3615125/should-wildcard-import-be-avoided
* https://www.reddit.com/r/learnpython/comments/8zej2c/eli5_what_are_wildcard_imports_and_why_should_i/

However I don't think we need to worry about it in this PR. Let's keep this PR small.

aswaterman:
I don't have a strong opinion on the matter, so I'll let the two of you hash this out, and will follow whatever is ultimately decided.

Timmmm:
@IIITM-Jay do you have any outstanding objections to this? As I understand it your comment was related to the `# from shared_utils import ...` comment which this PR doesn't touch.

IIITM-Jay:
@Timmmm No, I don't have any objections regarding the PR. Just a small thing, if possible, otherwise I could also take that up in a separate PR if you wish - could you please remove the wildcard imports and import only the necessary and sufficient methods of `shared` module, as I think it appears to be auto suggested with the methods of `shared` module while you were using reformatting tools, and then applied `#`, if it's not that much work from your end, otherwise I am ok to do that.

As for a single work we should not include two statements one with commented out and other as wildcard.

Let me know whatever suits you!

IIITM-Jay:
@Timmmm and @aswaterman I have removed the unwanted commented line for importing methods from `shared` module from the top and kept Wildacrd imports for now. Will raise a separate PR for importing all the required and necessary methods/ definition only from that module instead of wildcard imports.

aswaterman (approved):

PR #309, opened 2024-10-31
Add Unit Tests for shared_utils.py

TraXIcoN:
This PR adds unit tests for the shared_utils.py module. It covers encoding utilities, bit manipulation, argument handling, ISA type validation, string manipulation, and instruction processing. These tests will help ensure the module works correctly and is easier to maintain.

PR #310, opened 2024-10-31
Remove wildcard imports

Timmmm:
Use explicit imports rather than wildcards. This is more maintainable.

### DIFF
--- a/shared_utils.py
+++ b/shared_utils.py
@@ -8,7 +8,15 @@ import re
 from itertools import chain
 from typing import Dict, TypedDict
 
-from constants import *
+from constants import (
+    arg_lut,
+    fixed_ranges,
+    imported_regex,
+    overlapping_extensions,
+    overlapping_instructions,
+    pseudo_regex,
+    single_fixed,
+)
 
 LOG_FORMAT = "%(levelname)s:: %(message)s"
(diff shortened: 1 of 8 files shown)


shared_utils.py
```diff
@@ -16,30 +25,36 @@
 logging.basicConfig(level=LOG_LEVEL, format=LOG_FORMAT)
 
 
+def log_and_exit(message: str):
+    """
+    Log an error message and then exit with EXIT_FAILURE.
+    """
+    logging.error(message)
+    raise SystemExit(1)
+
```
IIITM-Jay, line 34: @Timmmm could you remove this as this is already been taken care as a part of #309 to fix `log_and_exit` things.
Thanks!

Timmmm: I will if that gets merged first, though maybe it would be better handled in #293 because I updated the file to use it everywhere in that PR. And it was first :-P

IIITM-Jay: Oh! I see, Thanks for letting me know. I was not not aware though.
Well in that case, I will resolve the conflicts/ update the PR with master branch for #309 , if in any case, as such arises.

IIITM-Jay:
@Timmmm I think these changes are already being merged as part of #311.

PR #311, opened 2024-11-04
Enable Pylint in CI and fix its errors

Timmmm:
The main fixes were:
    
* Specify encoding for all file opens. By default it depends on environment variables which is bad.
* Use `with` to open files. Otherwise they don't necessarily get closed.
    
There were also a few minor things like using `enumerate`, not using objects as default arguments, etc. In some cases I slightly refactored the code.

Note, this PR includes #310 because one of Pylint's checks is to not use wildcard imports. I suggest we merge that one first, then I'll rebase this.


aswaterman:
CI is failing, presumably because we need to pick a different version of pylint:

```
An unexpected error has occurred: CalledProcessError: command: ('/home/runner/.cache/pre-commit/repo7k6kzu03/py_env-python3/bin/python', '-mpip', 'install', '.')
return code: 1
stdout:
    Processing /home/runner/.cache/pre-commit/repo7k6kzu03
      Installing build dependencies: started
      Installing build dependencies: finished with status 'done'
      Getting requirements to build wheel: started
      Getting requirements to build wheel: finished with status 'done'
      Preparing metadata (pyproject.toml): started
      Preparing metadata (pyproject.toml): finished with status 'done'
    Collecting platformdirs>=2.2.0 (from pylint==3.3.1)
      Using cached platformdirs-4.3.6-py3-none-any.whl.metadata (11 kB)
    INFO: pip is looking at multiple versions of pylint to determine which version is compatible with other requirements. This could take a while.
stderr:
    ERROR: Ignored the following yanked versions: 2.13.0, 2.13.1
    ERROR: Ignored the following versions that require a different python version: 3.3.0 Requires-Python >=3.9.0; 3.3.1 Requires-Python >=3.9.0; 3.3.2 Requires-Python >=3.9.0; 3.3.3 Requires-Python >=3.9.0; 3.3.4 Requires-Python >=3.9.0; 3.3.5 Requires-Python >=3.9.0
    ERROR: Could not find a version that satisfies the requirement astroid<=3.4.0-dev0,>=3.3.4 (from pylint) (from versions: 1.0.0, 1.0.1, 1.1.0, 1.1.1, 1.2.0, 1.2.1, 1.3.0, 1.3.1, 1.3.2, 1.3.3, 1.3.4, 1.3.5, 1.3.6, 1.3.7, 1.3.8, 1.4.0, 1.4.1, 1.4.2, 1.4.3, 1.4.4, 1.4.5, 1.4.6, 1.4.7, 1.4.8, 1.4.9, 1.5.0, 1.5.1, 1.5.2, 1.5.3, 1.6.0, 1.6.1, 1.6.2, 1.6.3, 1.6.4, 1.6.5, 1.6.6, 2.0.0.dev0, 2.0.0.dev1, 2.0.0.dev2, 2.0.0.dev3, 2.0.0.dev4, 2.0, 2.0.1, 2.0.2, 2.0.3, 2.0.4, 2.1.0.dev0, 2.1.0, 2.2.0.dev0, 2.2.0.dev1, 2.2.0, 2.2.1, 2.2.2, 2.2.3, 2.2.4, 2.2.5, 2.3.0, 2.3.1, 2.3.2, 2.3.3, 2.4.0, 2.4.1, 2.4.2, 2.5, 2.5.1, 2.5.2, 2.5.3, 2.5.4, 2.5.5, 2.5.6, 2.5.7, 2.5.8, 2.6.0.dev0, 2.6.0, 2.6.1, 2.6.2, 2.6.3, 2.6.4, 2.6.5, 2.6.6, 2.7.0, 2.7.1, 2.7.2, 2.7.3, 2.8.0, 2.8.1, 2.8.2.dev0, 2.8.2, 2.8.3, 2.8.4, 2.8.5, 2.8.6, 2.9.0, 2.9.1, 2.9.2, 2.9.3, 2.10.0, 2.11.0, 2.11.1, 2.11.2, 2.11.3, 2.11.4, 2.11.5, 2.11.6, 2.11.7, 2.12.0, 2.12.1, 2.12.2, 2.12.3, 2.12.4, 2.12.5, 2.12.6, 2.12.7, 2.12.8, 2.12.9, 2.12.10, 2.12.11, 2.12.12, 2.12.13, 2.12.14, 2.13.2, 2.13.3, 2.13.4, 2.13.5, 2.14.0, 2.14.1, 2.14.2, 2.15.0, 2.15.1, 2.15.2, 2.15.3, 2.15.4, 2.15.5, 2.15.6, 2.15.7, 2.15.8, 3.0.0a1, 3.0.0a2, 3.0.0a3, 3.0.0a4, 3.0.0a5, 3.0.0a6, 3.0.0a7, 3.0.0a8, 3.0.0a9, 3.0.0b0, 3.0.0, 3.0.1, 3.0.2, 3.0.3, 3.1.0, 3.2.0, 3.2.1, 3.2.2, 3.2.3, 3.2.4)
    ERROR: No matching distribution found for astroid<=3.4.0-dev0,>=3.3.4
Check the log at /home/runner/.cache/pre-commit/pre-commit.log
```

IIITM-Jay:
> CI is failing, presumably because we need to pick a different version of pylint:


@aswaterman yes, you are absolutely correct here. Actually, as `pylint` **v3.3.1** requires `Python` **3.9** or later, which might be the cause I think. but as per the `Python` related versions, we already discussed in the #299 , regarding support of later versions and multiple versions as well.

aswaterman:
Now that I have merged #299, can you suggest to @Timmmm how best to proceed?

Timmmm:
Ah that's unfortunate. I think the simplest and best solution is just to skip Pylint on everything except the latest Python. There's no reason to run it on older versions because it doesn't use the current Python version to dictate what checks to run.

We can do the same for Pyright.

.pylintrc
```diff
@@ -0,0 +1,31 @@
+[MAIN]
+py-version = 3.6.0
```
IIITM-Jay, line 2: @Timmmm I think we could use `py-version` as `3.9.0`.

Timmmm: I think we want 3.6.0. According to [this](https://github.com/pylint-dev/pylint/issues/5038) it just hides suggestions for features that are only supported in newer versions of Python, so it should be set to the oldest version we support. In other words if we set it to 3.9.0 it will tell us to do some things that are not supported in 3.6.0, so we can't do them, like using the walrus operator for example.

Timmmm: Ah wait I didn't see the discussion about increasing the minimum version to 3.9 ( :tada: ). In that case I agree. I'll update the Pyright version too.

IIITM-Jay: Actually, as per the tagged PRs earlier, we would not be using the older versions of `Python` that are deprecated and reached its EOL. Hence, we are only supporting this repository to the latest and the supported versions of `Python`. 
And in #299 , we found that version  `python 3.6` is deprecated and reached its EOL, due to which we agreed not to support versions that are deprectaed by Python community itself.

IIITM-Jay:
@aswaterman I have configured the `pylint` to use version **3.9** as the `pylint` version **v3.3.1** requires `Python` **3.9** as I suggested @Timmmm too in my review comments.

I think the CI passes now.

aswaterman (approved):
Let’s run with it.

PR #312, opened 2024-11-05
Update the versions of all GitHub Actions

rpsene:
Update the versions of all GitHub Actions, as the old ones are using Node 12, which is deprecated.


IIITM-Jay (approved):
Thanks @rpsene !

@aswaterman As now we are already enabled the latest versions of `Python` for this repo, I think the updated versions of GitHub Actions also will be a better idea.

aswaterman (approved):

PR #315, opened 2024-11-05
Sort JSON output

Timmmm:
If we're sorting the dictionary anyway we may as well output the JSON sorted.

Also since Python 3.6 we can just use `dict()` since the insertion order is preserved.


aswaterman (approved):
OK by me.

PR #316, opened 2024-11-07
Restructured codebase

Myrausman:
Restructured the codebase by moving all opcode extension files into a single 'extensions' folder, and updated paths in the code to reflect this new structure.


shared_utils.py
```diff
@@ -457,7 +457,7 @@ def process_pseudo_instructions(
         logging.debug(f"Processing pseudo line: {line}")
         ext, orig_inst, pseudo_inst, line_content = pseudo_regex.findall(line)[0]
         ext_file = find_extension_file(ext, opcodes_dir)
-
+        # print("ext_file",ext_file)
```
aswaterman, line 460: Delete

shared_utils.py
```diff
@@ -514,6 +514,7 @@ def find_extension_file(ext: str, opcodes_dir: str):
         ext_file = f"{opcodes_dir}/unratified/{ext}"
         if not os.path.exists(ext_file):
             log_and_exit(f"Extension {ext} not found.")
+    # print(ext_file)
```
aswaterman, line 517: Delete

Makefile
```diff
@@ -1,4 +1,4 @@
-EXTENSIONS := "rv*" "unratified/rv*"
+EXTENSIONS := "/rv*" "unratified/rv*"
```
aswaterman, line 1: Confused why we need the `/` on the ratified extensions but not before the `unratified` directory?

aswaterman: Naively, I would've thought this would become `EXTENSIONS := "extensions/rv*" "extensions/unratified/rv*"`

IIITM-Jay: @aswaterman yes, you are correct, there is no need of putting a leading `/` here, and it will work correctly without adding it, @Myrausman you can verify the same.

IIITM-Jay: > Naively, I would've thought this would become EXTENSIONS := "extensions/rv*" "extensions/unratified/rv*"

@aswaterman but if we want to use explicit `extensions`, we have to probably use `EXTENSIONS := "../extensions/rv*" "../extensions/unratified/rv*"` using `relative path` as per the what `UNIX` like system says,

by adding `../` we are referring to the `extensions` folder by changing the present working directory `(pwd)` as `root` to `extensions` one, that's why required so.

IIITM-Jay: but yes I do agree with @aswaterman by explicit mentioning of `extensions` folder

IIITM-Jay: @Myrausman after whatever the changes being accepted, could you please verify also once, that whether the outputs that are generating with these changes are exactly matching with the existing structure, then we are all good to go!

Myrausman: > Naively, I would've thought this would become `EXTENSIONS := "extensions/rv*" "extensions/unratified/rv*"`

Yes, I tried it, and it wasn't producing any output in instr_dict.json. Therefore, I have proceeded with changes as follows: `EXTENSIONS := "../extensions/rv*" "../extensions/unratified/rv*"`

but the previous code as below was also working  and producing output:
`EXTENSIONS := "rv*" "unratified/rv*"`

IIITM-Jay:
@aswaterman as I know that I suggested for how to use `../extensions` but this also comes up with displaying in logger INFO as `Extensions selected` including path as well like below:
```
Running with args : ['./parse.py', '-c', '-go', '-chisel', '-sverilog', '-rust', '-latex', '-spinalhdl', '../extensions/rv*', '../extensions/unratified/rv*']
Extensions selected : ['../extensions/rv*', '../extensions/unratified/rv*']
```

But I think, the earlier is quite good in displaying as we are already stating on LHS as `Extensions selected` so again getting displayed as `../extension` is unnecessary. With existing one, it is more cleaner I guess, that selected extensions are `rv*` and the `unratified` ones like below:
```
Running with args : ['./parse.py', '-c', '-go', '-chisel', '-sverilog', '-rust', '-latex', '-spinalhdl', 'rv*', 'unratified/rv*']
Extensions selected : ['rv*', 'unratified/rv*']
```
 But again, whichever way you like, I think would be an appropriate one.

aswaterman:
I'll let you choose :)  Let me know when it's ready to merge.

IIITM-Jay:
@aswaterman I have kept the earlier version itself with no change in that, considering cleaner `logger INFO`  and also made changes to have optimized `pseudo` logic using conditional operator for a more concise approach.

I think it could be merged now.

aswaterman (approved):
SGTM.

PR #317, opened 2024-11-09
Speed-up the CI/ CD pipelines

IIITM-Jay:
### Description
This PR intends to speed-up the CI/CD workflows using various approaches.

### Approach Used

1. Used Cache to avoid reinstalling dependencies on every run.
2. Reduced redundant calls to pip install.


IIITM-Jay:
As some of the Opcodes Generation workflows taking longer than usual time for some python versions, closing this PR.

PR #318, opened 2024-11-09
Speed-up CI pipeline

IIITM-Jay:
### Description
This PR intends to speed-up the CI workflows using various approaches.

### Approach Used

1. Used Cache to avoid reinstalling dependencies on every run.
2. Reduced redundant calls to pip install.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -21,14 +21,19 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install Coverage
-      run: |
-        pip3 install coverage
+    - name: Cache pre-commit environment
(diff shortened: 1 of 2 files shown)


IIITM-Jay:
@aswaterman and @rpsene , we have gained the speed-up by these approaches as follows:

### Referenced PR (For comparison)
I am comparing with the last PR #316 having existing CI run as a reference, the total time it took for all versions to generate Opcodes Artifacts is **5m 3s** , that can be visualize as below:

image


### The Current PR (with Speed-Up)
The total time it took is **3m 49s** with speed-up approaches, thereby gaining a speed up of **1.24s**

image

aswaterman (approved):
Great!

PR #326, opened 2024-11-27
Refactored and Modularized C Artifact Generation

IIITM-Jay:
### Description
This PR intends to refactor the `c_utills` script using modularization.

### Approach

1. Modularize the code into smaller maintainable codes 
2. Used Argument and Return pattern for each function
3. The main function will only call the functions associated for each specific task

### DIFF
--- a/c_utils.py
+++ b/c_utils.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import pprint
+from typing import Dict, List, Tuple
 
 from constants import causes, csrs, csrs32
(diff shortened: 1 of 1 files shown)


IIITM-Jay:
@aswaterman This is in continuation with #283 . Will be modularizing and optimizing one file per PR, as it would be easier to review and quicker debugging as well if any so occurs.

aswaterman (approved):
Much cleaner!

PR #330, opened 2024-12-20
Switch Fence.tso to not pseudo

AFOliveira:
Just to get complient with the ISA Manual as clarified in https://github.com/riscv/riscv-isa-manual/issues/1782


AFOliveira:
This failed because they overlap (which I guess they should), should I go in the code and make this not an error?

aswaterman:
Or just leave it as a pseudo for pragmatic reasons.

AFOliveira:
> Or just leave it as a pseudo for pragmatic reasons.

Well, that would definitely be easier :), but can't this be problematic for users of the outputs? I realized that many outputs don't use pseudo, so couldn't it be misleading?

aswaterman:
The reason that it isn't problematic in practice is that it is deliberate that the FENCE encoding subsumes the FENCE.TSO encoding: the FENCE encoding is such that, if you ignore the bit that distinguishes FENCE.TSO from FENCE, the processor will still do the right thing (but in a slightly more conservative fashion).

So I don't want to spend time solving a problem that need not be solved.

AFOliveira:
I understand, thank you for the clarification @aswaterman.

aswaterman:
You're welcome, and thanks for playing ball.

PR #332, opened 2025-01-03
Simplify go's LUT

Jorropo:
I am working on C support and it started being very ackward to have to add things like funct4 and funct6 when the bits overlap and the some of the fields are used differently or not used at all by different encodings.

By passing the match the go assembler can do a simple bitwise or. This was also changed to emit go fmt-ed code from the start as go fmt now refuse to run due to the internal import.


aswaterman:
Can another user of the Go target look at this and provide feedback?

mengzhuo:
CC @4a6f656c @markdryan as Go riscv64 maintainers

Jorropo:
I would be surprising to have any user of the go target because it import private packages that are only found in the go compiler:
> ```go
> import "cmd/internal/obj"
> ```

4a6f656c:
> Can another user of the Go target look at this and provide feedback?

I expect that the only consumer is Go itself, however this needs a lot more discussion and review upstream, before a change of this nature is made.

PR #333, opened 2025-01-28
add hedelegh and mstateen0_priv113

JJ-Gaisler:
Needed for https://github.com/riscv-software-src/riscv-isa-sim/pull/1896


aswaterman (approved):

aswaterman:
Thanks.

Issue #339, opened 2025-03-05
[proposal] add reference sail implementation for each instruction


sequencer:
In our downstream repository:
https://github.com/chipsalliance/rvdecoderdb/pull/39
We are experimenting with adding a sail code snippet for each single instruction, for example: 
https://github.com/rez5427/rvdecoderdb/blob/sailcodegen/rvdecoderdbtest/jvm/src/sail/inst/rv_i/addi
We are wondering would it be possible for riscv-opcodes accepting this try as the upstream?
I personally think this can be used for representing the details semantics of each instructions but with some limitations:
For instructions w/o architecture state changes, e.g wfi, they should be left empty with an external C-API call. 
For institutions without a deterministic execution result, a reference implementation would be added but can be overridden by downstream user, e.g. vector do reduce add.
For instructions leading to change the memory state, e.g load store, the implementation should have the C-API inside which for avoiding concrete implementation.

The codegen emulator can be maintained in the downstream repositories to save the maintenance burden for this project.

PR #341, opened 2025-03-08
Add sail implementation

rez5427:
relate to https://github.com/riscv/riscv-opcodes/issues/339

### DIFF
--- a/sail/inst/rv_c/c_addi4spn
+++ b/sail/inst/rv_c/c_addi4spn
@@ -1,0 +1,3 @@
+let imm : bits(12) = (0b00 @ nzimm @ 0b00);
+let rd = creg2reg_idx(rdc);
+X(rd) = X(sp) + imm;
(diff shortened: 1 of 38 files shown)


sequencer:
Here are some of my concerns:
- How to define the architecture states, and the access API of them, they should have a concrete documentation or even implementation?(I personally prefer documentation).
- Not all architecture states are necessary for different extensions, so the problem becomes: how to define the architecture state for each extension firstly. And of cause please provide corresponding ABI for accessing each computer architecture states.
- CSR is tricky, maybe you need to add more CSR regfileds info to riscv-opcodes firstly.
- How to use it? Should a codegen-based emulator living in this repo? I don't this so, but it can be added for generating the pdf specifications.

Another point is we already have riscv/sail-riscv, and we should add these codes their. However from my point of view, that's only a concrete RISC-V software implementation, doesn't formally define the instruction and the architecture state, we need to have a sound definition of extensions and each instructions.

But after all, the repo is riscv-opcodes, which only defines the opcode for RISC-V, I personally like this, but anyway need to ask @aswaterman to decide should we add these sail implementation snippets here:)

PR #355, opened 2025-06-13
Remove pseudo_op to rv64_i for slli,srli,srai

Avimitin:
Instructions slli,srli,srai have different bit pattern in rv32_i versus rv64_i. It is better to re-define them, and alias those `_rv32` suffix instruction to rv32_i self specification.

### DIFF
--- a/extensions/rv32_i
+++ b/extensions/rv32_i
@@ -1,6 +1,6 @@
-$pseudo_op rv64_i::slli slli rd rs1 shamtw 31..25=0  14..12=1 6..2=0x04 1..0=3
-$pseudo_op rv64_i::srli srli rd rs1 shamtw 31..25=0  14..12=5 6..2=0x04 1..0=3
-$pseudo_op rv64_i::srai srai rd rs1 shamtw 31..25=32 14..12=5 6..2=0x04 1..0=3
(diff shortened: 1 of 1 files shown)


sequencer:
Based on the README, the semantic of `pseudo` is:
> These are instructions which are aliases of regular instructions. Their encodings force certain restrictions over the regular instruction.

It means: the `pseduo` instruction can be covered by the original instruction.
For example, `jalr` can cover the `ret` encoding space, by constrains `rs1` and `imm` to 0. 
Generally, the `pseduo` instruction alias can provide to the hardware decoder with additional information for micro-architecture: for example, `ret` is special to pop from the RAS in the BPU.

So does the `?_rv32` mean a `pseduo`? I think it is, based on the `pseduo` semantic.  
However, in this case, the original and pseudo instructions cannot live together, although the 32-bits version of these instructions are covered by 64-bits.

I think the true issue is the existence of the instructions with `_rv32` suffix, which is actually the "old name" like [`fmv.x.s`](https://github.com/riscv/riscv-opcodes/blob/9a09c4844f8eb926e4d4f96c6adc097b78d8b945/extensions/rv_f#L28-L30).

PR #360, opened 2025-07-18
Move Zilsd & Zclsd to ratified extensions directory

christian-herber-nxp:


aswaterman:
@christian-herber-nxp you should be able to view the CI reports, but FWIW, here it is:

```
ERROR:: Instruction ld from rv32_zilsd has different encodings in different base ISAs
Extensions selected : ['rv*', 'unratified/rv*']
Error: Process completed with exit code 1.
```

I'll leave it to you to decide how best to fix it.

christian-herber-nxp:
I have seen it, but will have to most likely push this until after my PTO. It could be I get this done next week, otherwise a couple weeks later.

christian-herber-nxp:
@aswaterman : The issue that cause the failure is fixed.
However, I am not sure this is the best way to solve it.
The instr dict will only ever contain one version of ld (RV64I or Zilsd).
Also, both ld/sd variants must have the exact same encoding.
Again, not sure if that is correct, because technically speaking, the LSB of the pair register is fixed to zero, but appears programmable in the encoding.
If the dict had a RV32 dict and RV64 dict inside, for the base ISA specific instructions, it would be possible to do.

aswaterman:
I think this case is analogous to `slli` etc., where bit 5 of the shift amount must be 0 in RV32.  In that case, we defined the instructions for RV64 (using the 6-bit `shamt` field), but for RV32, we only defined pseudos (using the 5-bit `shamtw` field).

christian-herber-nxp:
> I think this case is analogous to e.g. `slli`, where bit 5 of the shift amount must be 0 in RV32.  In that case, we defined the instruction for RV64 (using the 6-bit `shamt` field), but for RV32, we only defined a `$pseudo_op` (using the 5-bit `shamtw` field).

But slli is actually the same behavior. In the zilsd case, it is somewhat different. 
I think we also need to look forward to P. 
There, iirc, the lab of pair registers is used to encode a func field?

christian-herber-nxp:
@aswaterman : Made the change to pseudo op. Should now be ok.

aswaterman (approved):
Thanks for playing ball, @christian-herber-nxp!

PR #369, opened 2025-09-23
Add P extension

2011eric:
This commit add encodings for P Extension:
https://www.jhauser.us/RISCV/ext-P/RVP-instrEncodings-015.pdf


2011eric:
@chihminchao 
Thanks for reviewing.
For now I mark them as overlapping extensions in `constraint.py`. 
And I do think it is necessary to include them in this repo due to better alignment with the spec.
I think I can mark them as `pseudo_op` to prevent the emition of their masks, or I can simply comment them out.

aswaterman:
Thanks for submitting this PR.  I have heard that there will be another round of renaming these instructions before the spec is submitted for architecture review.  I think we should wait until that occurs, and leave this PR open in the meantime.

chihminchao:
It could be closed since most of work has be included in https://github.com/riscv/riscv-opcodes/pull/409.

PR #379, opened 2025-10-15
Making matplotlib a dynamic dependency for svg generation

christian-herber-nxp:


aswaterman:
@christian-herber-nxp I already did this. The version on master only relies on matplotlib if -svg is passed.

christian-herber-nxp:
> @christian-herber-nxp I already did this. The version on master only relies on matplotlib if -svg is passed.

did not see that.

Issue #382, opened 2025-10-29
Splitting rv_zabha into rv_zabha and rv_zabha_zacas


robinali-codasip:
Currently `extensions/rv_zabha` contain the `amocas.h` and `amocas.b` instructions, but these are only present if `Zacas` is also implemented.

Shouldn't the mentioned instructions be placed into their own file `rv_zabha_zacas`?

PR #384, opened 2025-10-29
Split zabha into zabha and zabha_zacas

robinali-codasip:
* Move AMOCAS.H and AMOCAS.B instructions into the new rv_zabha_zacas file

Related #382


robinali-codasip:
@aswaterman
Should be fixed, actually ran pre-commit this time :sweat_smile:

PR #386, opened 2025-11-05
ext: smpmpmpt: new field for pmp config

chihminchao:
There are two reserved bits used to controll non-cacheable and io attribute for pmp regsions

ref
  1. https://riscv.atlassian.net/browse/RVS-2948
  2. https://github.com/riscv/riscv-isa-manual/blob/smpmpmt/src/smpmpmt.adoc

### DIFF
--- a/encoding.h
+++ b/encoding.h
@@ -366,6 +366,7 @@
 #define PMP_W     0x02
 #define PMP_X     0x04
 #define PMP_A     0x18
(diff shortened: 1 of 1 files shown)


aswaterman (approved):

Stan-Cho:
The field name `PMP_A` seems to be a duplicate. I think it should be `PMP_MT`.

PR #388, opened 2025-11-07
ext: spmp (1.0.0-rc2): new PMP entry field

binno:
These two fields are defined separately for PMP entry in SPMP extension.

* Reference
   1. rv-spmp-spce 1.0.0-rc2
       https://github.com/riscv/riscv-spmp/tree/main


aswaterman (approved):

chihminchao:
I think it is based on 1.0.0-rc2.  It is better to describe it in commit log

PR #389, opened 2025-11-13
Add Supervisor Domains extension and CSRs

rsahita:
Add Supervisor Domains extension and CSRs


encoding.h
```diff
@@ -154,17 +154,34 @@
...
+#define MIE_STIE            (1 << IRQ_S_TIMER)
+#define MIE_VSTIE           (1 << IRQ_VS_TIMER)
+#define MIE_MTIE            (1 << IRQ_M_TIMER)
+#define MIE_SEIE            (1 << IRQ_S_EXT)
+#define MIE_VSEIE           (1 << IRQ_VS_EXT)
+#define MIE_MEIE            (1 << IRQ_M_EXT)
+#define MIE_SGEIE           (1 << IRQ_S_GEXT)
+#define MIE_LCOFIE          (1 << IRQ_LCOF)
+#define MIE_MSDEIE          (1 << IRQ_MSDEI)
+
```
aswaterman: Let's get rid of this.  It is a property of the ISA that MIP and MIE bits line up with each other, so having a redundant set of macros is just more code to maintain.

aswaterman (commented):
@ved-rivos can you also take a quick look?  Do you agree this is stable enough that it's safe to merge?

aswaterman:
Thanks @rsahita!

pathakraul:
For the development of SMMTT support in Qemu and OpenSBI its imperative to have some confirmation on the proposed encodings of mfence/minval instructions and other CSRs and fields.

Though riscv-smmtt spec is not yet frozen which may be gating this PR but please let us know if atleast SMSDID and SMMPT introduced instructions and CSRs encodings are good enough in the PR.

aswaterman:
One of the spec authors will have to chime in and make sure that this PR matches the most recent version of the spec.

pathakraul:
@rsahita Can you please confirm?

PR #392, opened 2025-11-19
go: make package comment more ergonomic

mengzhuo:
Go use explict extension, as extensions grows
it's harder for developer to add extentions without change the whole generated line.
This PR generate right aligned comments and split lines by four sorted extensions
i.e.
```
make inst.go EXTENSIONS='\
   rv64_a rv64_c rv64_d rv64_f \
 rv64_i rv64_m rv64_q rv64_zba \
   rv64_zbb rv64_zbs rv_a rv_c \
         rv_c_d rv_d rv_f rv_i \
      rv_m rv_q rv_s rv_system \
     rv_v rv_zba rv_zbb rv_zbs \
            rv_zicond rv_zicsr \
'
```
So that developers can add more extensions by copy/paste on terminal without pain.


src/riscv_opcodes/go_utils.py
```diff
@@ -9,12 +8,19 @@
...
+def make_go(instr_dict: InstrDict, extensions: list[str]):
+    comment = """// Code generated with following arguments; DO NOT EDIT.
+/*
+make inst.go"""
+    if extensions:
+        comment += " EXTENSIONS='\\\n"
+        n = 4
+        exts = sorted(extensions)
+        for i in range(0, len(exts), 4):
+            ext_line = " ".join(exts[i : i + n])
```
IIITM-Jay: @mengzhuo , This is a bit misleading. The piece of code will end up in breaking the logic intended. The chunk size is controlled by `n` and the iteration is hardcoded as `4` and again inner logic still depends on n. If someone later changes `n` to `8` _(thinking it controls grouping)_, the code breaks because, outer loop still advances by `4`, but inner slice chooses `8`. Consequently, creating duplicated extensions and overlapping groups, leading to generating incorrect outputs.

src/riscv_opcodes/go_utils.py
```diff
@@ -9,12 +8,19 @@
...
+            comment += f"{ext_line:>30} \\\n"
```
IIITM-Jay: IMO, it would be better to have adjusting logic to support variable-length, instead of making it as a hardcoded limit

src/riscv_opcodes/go_utils.py
```diff
@@ -9,12 +8,19 @@
...
-
-    prelude += """
+def make_go(instr_dict: InstrDict, extensions: list[str]):
+    comment = """// Code generated with following arguments; DO NOT EDIT.
+/*
+make inst.go"""
+    if extensions:
+        comment += " EXTENSIONS='\\\n"
+        n = 4
+        exts = sorted(extensions)
```
IIITM-Jay: I am not sure with the sorting of extension here as if suppose a tool relies on a particular order _(for readability, or minimal diff when adding one extension)_, this change may cause larger diffs than necessary each time a new extension is added.

src/riscv_opcodes/go_utils.py
```diff
@@ -9,12 +8,19 @@
...
-    args = " ".join(sys.argv)
-    prelude = f"""// Code generated by {args}; DO NOT EDIT."""
-
-    prelude += """
+def make_go(instr_dict: InstrDict, extensions: list[str]):
+    comment = """// Code generated with following arguments; DO NOT EDIT.
+/*
+make inst.go"""
+    if extensions:
+        comment += " EXTENSIONS='\\\n"
```
IIITM-Jay: As these are going to be inside comment `/*...*/`, I believe that the backslashes have no effect on Go compilation.

IIITM-Jay (commented):
@mengzhuo Thanks for the  PR!

I have left some comments which I felt to be addressed. Feel free to keep your opinions as well.

PR #399, opened 2026-01-05
Make Rust constants public

Timmmm:
This is the more flexible option. It's easy to make them all private again by just not making the module itself public. But as it is making them all public requires hacky regex replacements in the file.


aswaterman (approved):
Makes sense to me, but I'll let this marinate for a while so some of the Rust users can chime in.  Ping me again in a few days if there's no opposition and I'll merge it.

PR #407, opened 2026-02-19
Add CSR_MEDELEGH

Zain2050:


aswaterman (commented):
Can you add `hedelegh` (0x612) as part of this PR, please?

Zain2050:
hedelegh is already added, but medelegh was missing.

csrs.csv
jordancarlin: Shouldn’t this be added to `csrs32.csv` instead?

aswaterman (approved):

PR #408, opened 2026-03-02
Fix unset variable in Makefile

jmgorius:
Use `uv run` as the command runner if `uv` is available.


aswaterman (approved):
Oops, not sure what I was thinking.

PR #411, opened 2026-03-12
fix absolute path for inst.go codegen

mengzhuo:
sys.args will print out absolute path like
"/x/y/z -go ...", this will change if path is different.

This PR replace the absolute path with
"uv run riscv_opcodes" for stable code generation.


aswaterman:
This repo is designed to work without requiring `uv` (see the `Makefile` to see what I mean).  Please fix this PR to not require `uv`, and test it accordingly, and ping me again.

mengzhuo:
Hi @aswaterman, I had fixed it, PTAL

aswaterman (approved):

PR #413, opened 2026-03-24
ext-p : update table to v0.20

chihminchao:
Quoted from spec

"This version (020) differs from the previous one (019) with the addition of:
  * SHL instructions (shift left or right, logical) that are unsigned versions of the SHA instructions (shift left or right, arithmetic); and
  * for RV64 only, ‘P’-suffi x versions of PNCLIP instructions that are similar to RV32 PNCLIP instructions without shifting or rounding.

The complete set of instructions added for RV32 are:
  PSSHL.HS SSHL PSSHL.DHS PSSHL.DWS
  PSSHLR.HS SSHLR PSSHLR.DHS PSSHLR.DWS
And for RV64:
  PSSHL.HS PSSHL.WS SHL PNCLIPP.B PNCLIPP.H PNCLIPP.W
  PSSHLR.HS PSSHLR.WS SHLR PNCLIPUP.B PNCLIPUP.H PNCLIPUP.W"

reference:
  https://www.jhauser.us/RISCV/ext-P/RVP-instrEncodings-020.pdf

### DIFF
--- a/extensions/unratified/rv32_p
+++ b/extensions/unratified/rv32_p
@@ -10,6 +10,8 @@ $pseudo_op rv64_p::psslai.w sslai                 31=1 30..28=0x5 27=0 26..25=0x
 
 
 # Page 7: Scalar Operand Shift Instructions - Pseudo-instructions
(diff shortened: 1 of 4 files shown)


aswaterman (approved):
LGTM, thanks for the quick turnaround!

PR #414, opened 2026-03-25
ext-p: fix typo

chihminchao:
The opcode[1:0] should 3 but not 2 for 32bits encoding


chihminchao:
It should be caused by Ctrl-a + Ctrl-a in vim

chihminchao:
@aswaterman The part is included in https://github.com/riscv-software-src/riscv-isa-sim/pull/2246

aswaterman (approved):
> It should be caused by Ctrl-a + Ctrl-a in vim

Happens to the best of us :)

Issue #416, opened 2026-04-09
rv_zimop and rv_zcmop is inconsistent with other extensions


robinali-codasip:
The `rv_zimop` and `rv_zcmop` files use uppercase characters in the instruction mneumonics (e.g `mop.r.N`) and have oddly named operands

In addition, personally the pseudo approach here is odd - the instructions are `mop.r.0` - `mop.r.31`. Which would remove the extra defined operand and also allow new extensions that use the space to use $pseudo easier.

PR #417, opened 2026-04-09
rv_zimop/rv_zcmop - rename mneumonics to use all lowercase & simplify operand names

robinali-codasip:
* mop.r{r}.N -> mop.r{r}.n to keep consistency
* mop_r{r}_t -> mop_r{r}
Same for c.mop.n

Issue #416  - Change from pseudo is not implemented here so favour for that would be appreciated here


aswaterman (commented):
This is a backwards-incompatible change for downstream users, and it isn't clear to me that it's necessary.  It is true that the _N_ is unusual, but there is an arguable justification for that: there is no such instruction mnemonic as `c.mop.N`, as the _N_ is a placeholder for an integer in the mnemonic.

robinali-codasip:
@aswaterman I agree that it's not ideal for backwards compatibility, but I am not convinced myself that improvements can be ignored in favour of maintaining it. Would you be more likely to accept a change if I also removed the pseudo definitions and make each instruction its own entry? i.e `c.mop.0`, `c.mop.1` etc. (removing `c.mop.N`). I believe that it makes it better for future extensions that use the _mop_ space.

aswaterman:
I guess I'm not sure I agree it's an improvement.  What's the problem, exactly, aside from an apparent inconsistency?  (And as I wrote above, I still think the inconsistency is justified since N is a placeholder for an integer, rather than part of an official mnemonic.)

Getting rid of `c.mop.N` will be more disruptive than renaming it to `c.mop.n`, FWIW.

robinali-codasip:
I understand that getting rid of `c.mop.N` altogether will be more disruptive, it's what I think is really the issue here. It obfuscates what the instructions and mneumonics are, for example adding a new instruction that uses the mop space will need to create an identical pseudo to an existing entry here instead of a new pseudo which maps one-to-one. I think it's a bit like having one entry for `addsub` and then `add` and `sub` pseudos. The idea behind suggesting the new change instead is that it's a full solution rather than a half solution currently in the current PR. I do now agree that `c.mop.N` -> `c.mop.n` is probably not worth the disruption. But I do not right now for the full change.

aswaterman:
OK, well, I'm on record not supporting that change.

PR #418, opened 2026-04-30
Allow overlapping extensions with `--warn-overlap`

colluca:
This PR loosens the behaviour of the parser in the presence of overlapping instruction. Instead of failing, if the `--warn-overlap` flag is passed, warnings are reported in the presence of instructions with overlapping encodings.

This feature is needed to e.g. generate a single `inst.sverilog` file with all instructions for a core that can support multiple custom ISA extensions (possibly overlapping) at configuration time.

This is the case for the Snitch core for example. Though overlapping ISA extensions cannot be simultaneously enabled, it can be configured to support different (overlapping) ISA extensions at configuration time.
This is currently achieved by generating a single `inst.sverilog` file, and using a single case statement to implement the decoder, with conditionals to select overlapping instructions, e.g.:
```systemverilog
unique casez (inst_data)
  P_SH_IRPOST,
  SCFGRI: begin
    if (Xpulppostmod) begin
      // Decode P_SH_IRPOST instruction in Xpulppostmod ISA extension
    end else if (Xssr) begin
      // Decode SCFGRI instruction in Xssr ISA extension
    end
  end
endcase
```

### DIFF
--- a/src/riscv_opcodes/parse.py
+++ b/src/riscv_opcodes/parse.py
@@ -31,8 +31,9 @@ def generate_extensions(
     go: bool,
     latex: bool,
     svg: bool,
+    warn_overlap: bool = False,
 ):
-    instr_dict = create_inst_dict(extensions, include_pseudo)
+    instr_dict = create_inst_dict(extensions, include_pseudo, warn_overlap=warn_overlap)
     instr_dict = dict(sorted(instr_dict.items()))
     instr_dict_with_segment = add_segmented_vls_insn(instr_dict)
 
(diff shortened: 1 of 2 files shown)


aswaterman (approved):
OK with me.

aswaterman:
The code formatter seems to be choking on this PR; ping me when it's ready to merge.

colluca:
Thanks for the review. I added a commit that should fix the formatting errors.

aswaterman:
Thanks!

PR #422, opened 2026-06-22
Mirror ISA extension organization for CSRs

colluca:
To support managing custom (and possibly conflicting) sets of CSRs, we propose to reorganize the CSR definitions, mirroring the organization of the ISA extensions.

Ratified CSR sets would live in the `csrs/` folder, while custom CSR sets go in `csrs/unratified`.
The list of desired CSR sets can be passed to the CLI, similar to how the CLI accepts a list of ISA extensions.
To this end, a new `--csr` flag is provided.

### Usage

To add a custom CSR set, place a file under `csrs/unratified/`:

```csv
# MyVendor CSRs
0x800, mv_ctrl
0x801, mv_status
```
Then pass `--csr rv_myvendor` to include only that set, or omit `--csr` to include all CSR sets, including ratified sets.

### Changes

- `read_lines` is moved to `resources.py` so it can be reused also in `constants.py`, and in general by any file that needs to read some resource
- `read_int_map_csv` is updated to use `read_lines` instead of opening the CSV resource directly, enabling #-prefixed comment lines and blank lines in all CSV files, as supported in ISA extension files
- the module-level `csrs` and `csrs32` globals are removed entirely. Generators no longer import CSR data from `constants` but instead take a CSR dictionary as argument, which is populated with the CSR sets selected on the CLI (similar to how instructions are handled)
- `csrs.csv` and `csrs32.csv` have been moved to the new `csrs/` subdirectory
- a `--csr` CLI argument (repeatable) is added to specify which CSR sets to load. Omitting `--csr` includes all sets. `generate_extensions` gains a `csrs` parameter; it calls `create_csr_dict` once and passes the resulting `csr_dict` to every generator.
- all generators gain a `csr_dict: CsrDict` parameter. Most generators are agnostic to which sets are present. The exception are Chisel's `all` and `all32` arrays, which retain their semantics by reading `csr_dict.get("csrs", [])` and `csr_dict.get("csrs32", [])` respectively.


aswaterman:
Seems reasonable at first blush, but I'm hoping someone else has some time to do a code review.

PR #426, opened 2026-07-28
Add Zilx indexed integer load instruction encodings

wangpc-pp:
Add the `Zilx` family of indexed integer load instructions (v0.1), which reuse the AMO major opcode in an R-type form where rs1 is the index and rs2 is the base. funct5 selects the addressing mode (unscaled, scaled, or scaled with a zero-extended 32-bit index) and funct3 reuses the base integer load width/signedness encoding, with `aq` and `rl` held at 0.

Shared RV32/RV64 forms (`lxh/lxw/lxhu`, the scaled `lxs*` loads, and the `lxb/lxbu` pseudoinstructions for the scaled byte loads) live in `rv_zilx`. RV64-only forms live in `rv64_zilx`: the doubleword and unsigned-word loads plus the scaled unsigned-word-index (`lxsuw*`) mode.

Doc: https://github.com/riscv/riscv-zilx/blob/main/src/unpriv/zilx.adoc

### DIFF
--- a/extensions/unratified/rv64_zilx
+++ b/extensions/unratified/rv64_zilx
@@ -1,0 +1,25 @@
+# rv64_zilx
+# Zilx indexed integer load instructions available only on RV64:
+#   - doubleword loads (funct3=011)
+#   - unsigned-word loads (funct3=110)
+#   - the scaled unsigned-word-index mode (funct5=11110), which zero-extends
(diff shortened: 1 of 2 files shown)


wangpc-pp:
Ping. A lot of downstream softwares like Spike, Golang are depending on this.

aswaterman (approved):
Thanks for the ping.

PR #427, opened 2026-07-29
Add xCAUSE interrupt and exception code field masks

colluca:
## Summary
- Adds `MCAUSE32/64_INT` and `MCAUSE32/64_EXCCODE` masks for the `mcause` Interrupt and Exception Code fields, per the RISC-V privileged spec.
- Adds XLEN-generic `MCAUSE_INT`/`MCAUSE_EXCCODE` aliases in the `__riscv_xlen` selection block, following the same pattern already used for `MSTATUS_SD` and `SATP_MODE`.
- Adds `SCAUSE_*`/`UCAUSE_*` equivalents, since `scause`/`ucause` share the identical bit layout as `mcause`.


aswaterman (commented):
There is no `ucause` CSR, so we should remove those macros, but the rest of this LGTM.

colluca:
Thank you for the prompt review.

> There is no `ucause` CSR, so we should remove those macros, but the rest of this LGTM.

Good point, I fixed that now.

aswaterman (approved):

PR #430, opened 2026-08-20
update opcode for zvabd

keyon1992:
ARC assigned new opcodes for Zvabd extension, we need to update it.
see https://github.com/riscv/riscv-isa-manual/pull/3279

### DIFF
--- a/extensions/unratified/rv_zvabd
+++ b/extensions/unratified/rv_zvabd
@@ -1,6 +1,9 @@
 # rv_zvabd
-vabs.v        31..26=0x12 vm vs2 19..15=0x10 14..12=0x2 vd 6..0=0x57
-vabd.vv       31..26=0x11 vm vs2 vs1 14..12=0x2 vd 6..0=0x57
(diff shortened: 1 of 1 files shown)


keyon1992:
@aswaterman Could you please review it

aswaterman (approved):

peter-bergner:
I think the *.vx versions of the insns have a typo, don't they?  For example:

`+ vabd.vx       31..26=0x15 vm vs2 vs1 14..12=0x6 vd 6..0=0x57`

...lists source operands of vs1 and vs2.  Shouldn't one of those be rsN rather than vsN?

keyon1992:
> I think the *.vx versions of the insns have a typo, don't they? For example:
> 
> `+ vabd.vx 31..26=0x15 vm vs2 vs1 14..12=0x6 vd 6..0=0x57`
> 
> ...lists source operands of vs1 and vs2. Shouldn't one of those be rsN rather than vsN?

thanks, I‘ill fix it soon

PR #431, opened 2026-08-23
Zalasr is ratified

aswaterman:


jordancarlin:
Convenient timing. We just ran into these being missing in the generated files for riscv-arch-test (looks like we are only generating using the ratified encodings).

aswaterman:
@jordancarlin not a concidence, it turns out; @adingank-qualcomm asked me about it out-of-band.

With that said, most downstream tools should feel free to use stuff in the `unratified` directory.  riscv-arch-test should do the same.  It's a race to the bottom if nothing can be merged anywhere until stuff is actually ratified.

jordancarlin:
> With that said, most downstream tools should feel free to use stuff in the `unratified` directory. riscv-arch-test should do the same. It's a race to the bottom if nothing can be merged anywhere until stuff is actually ratified.

Agreed. I didn't realize we were only using the ratified encodings until seeing this PR. Planning to change that the next time I update the file.

PR #433, opened 2026-09-07
Fix wrong syntax for spinalhdl generation

inochisa:
In scala, "->" is usually used in map. For function "def", it is more common to use "=" for definition, and using "->" is an invalid syntax.


aswaterman (approved):
> In scala, "->" is usually used in map. For function "def", it is more common to use "=" for definition, and using "->" is an invalid syntax.

Sure, but you're the one who contributed this code in the first place...

PR #437, opened 2026-09-18
Fix operand metadata for dot-product instructions

zeyi2:
The Zvbdota specification assigns `vs2[2:0]` to the scaled `ci` immediate, leaving `vs2[4:3]` to select the `EMUL=8` source register group.

This commit splits the existing `vs2` field into `vs2_bdot` and `ci`. This follows the existing way in LLVM: https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/RISCV/RISCVInstrInfoZvbdota.td#L31-L37

Spec: https://github.com/riscv/riscv-isa-manual/blob/c152059fd5db62fe5fe2b3e302bdfa130306e41c/src/zvdota.adoc#L168-L173


aswaterman (commented):
`vs2x8` instead of `vs2_bdot`?  (The design pattern is potentially more general.)

zeyi2:
> `vs2x8` instead of `vs2_bdot`? (The design pattern is potentially more general.)

Fixed, thanks!

aswaterman (approved):

